PR #2, opened 2020-03-15
mountinfo: add go.mod

kolyshkin:
This module is not dependent on any other modules, so it needs to be
added first.

Once added, we can do the same for `moby/sys/mount`

### DIFF
--- a/mountinfo/go.mod
+++ b/mountinfo/go.mod
@@ -1,0 +1,3 @@
+module github.com/moby/sys/mountinfo
+
+go 1.14


kolyshkin:
Merging this myself since this is super trivial, and it should unblock me from "modularization" of moby/sys/mountinfo, and then adding some CI

cpuguy83:
I'm not sure why we needed this module. Multi-module is pretty rough, especially for minimal dependencies as is.

PR #3, opened 2020-03-15
mount: add go.mod

kolyshkin:
The version of golang.org/x/sys corresponds to the one currently used
by Moby.

Once there, it needs to be tagged, e.g.
```
git tag mount/v0.1.0
```

### DIFF
--- a/mount/go.mod
+++ b/mount/go.mod
@@ -1,0 +1,8 @@
+module github.com/moby/sys/mount
+
+go 1.14
+
+require (
+	github.com/moby/sys/mountinfo v0.1.0
+	golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae
+)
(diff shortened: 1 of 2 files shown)


kolyshkin:
@thaJeztah PTAL

thaJeztah:
are we sure we want mount and mountinfo to be separate modules, and to be tagged separately? (wondering how much they will evolve separately)

kolyshkin:
> are we sure we want mount and mountinfo to be separate modules, and to be tagged separately? (wondering how much they will evolve separately)

The main reason to separate it was that `mountinfo` is not dependent on anything (except go std lib), while mount depends on golang.org/x/sys/unix and `mountinfo`. So, if someone wants to use mountinto only, they don't have to pull any extra deps.

The reason for separate tagging is this is how you do it in a repo with multiple modules (https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories). This also allows to bump a single module without affecting others.

kolyshkin:
Merging to unblock tagging and CI

PR #4, opened 2020-03-15
mount: test nitpicks

kolyshkin:
### mount: fix test for added relatime
    
The mount option `relatime` can be added by the kernel, and
this is what happens on my machine:
    
> --- FAIL: TestMount (0.00s)
>     mounter_linux_test.go:176: unexpected mount option "relatime", expected ""
> FAIL
    
Same as in commit afc7b0dd970, this option should be ignored.

##  mount: test nitpicks
    
1. Do not ignore errors from Mkdir.
2. Use ioutil.WriteFile for simplicity, to not leak fds and to not
       ignore errors from Write or Close.

### DIFF
--- a/mount/mount_unix_test.go
+++ b/mount/mount_unix_test.go
@@ -3,6 +3,7 @@
 package mount
 
 import (
+	"io/ioutil"
 	"os"
 	"path"
 	"testing"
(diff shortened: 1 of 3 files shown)


kolyshkin:
@cpuguy83 @thaJeztah PTAL

thaJeztah:
could you add CI so that tests actually run on this repo? probably could just copy what's in https://github.com/moby/term/blob/master/.github/workflows/test.yml

kolyshkin:
I plan to add CI once #3 is merged (don't remember ATM why it's needed but it's needed)

kolyshkin:
merging to unblock adding CI

PR #6, opened 2020-03-17
mountinfo: add GetMountsFromReader

kolyshkin:
Some code have tests or benchmarks that come with fake mountinfo.
To switch those to use moby/sys/mountinfo without any major rework
of those tests and benchmarks we need to provide a way to work
with such fake mountinfo.

This is probably more important for benchmarks, since without
this we can't make any comparisons after switching to moby/sys.

### DIFF
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -1,9 +1,19 @@
 package mountinfo
 
+import "io"
+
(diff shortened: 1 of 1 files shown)


kolyshkin:
rebased to kick CI

kolyshkin:
Windows CI failures are expected (and are fixed in #8)

PR #8, opened 2020-03-17
mount: separate mount error wrapper

kolyshkin:
This is only used for !windows, so move this to a separate
source file to avoid the following windows linter warning:

> ##[error]mount.go:14:6: type `mountError` is unused (unused)

This is just moving stuff around, no functional change


kolyshkin:
Yay, CI! :fireworks: @thaJeztah PTAL

PR #10, opened 2020-03-20
mountinfo: fix windows build

kolyshkin:
When compiling for windows, we get this error:

> GOOS="windows" GOARCH="amd64" GOARM=""
> vendor/github.com/moby/sys/mountinfo/mountinfo.go:16:9: undefined: parseInfoFile

The problem is newly added GetMountsFromReader() (see https://github.com/moby/sys/pull/6)

Amend the `_windows.go` and `_unsupported.go` to fix.

NOTE that eventually we need to deprecate windows support for this
library, since it's not really doing anything.

### DIFF
--- a/mountinfo/mountinfo_unsupported.go
+++ b/mountinfo/mountinfo_unsupported.go
@@ -4,9 +4,14 @@ package mountinfo
 
 import (
 	"fmt"
+	"io"
 	"runtime"
 )
 
-func parseMountTable(f FilterFunc) ([]*Info, error) {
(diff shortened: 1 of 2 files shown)


kolyshkin:
Found in https://github.com/moby/moby/pull/40694#issuecomment-601415142

kolyshkin:
> LGTM, but curious why CI didn't catch this

Turns out that it did (see https://github.com/moby/sys/runs/514928353?check_suite_focus=true) but we overlooked it because some were expected (as per https://github.com/moby/sys/pull/8)

PR #11, opened 2020-03-30
Makefile: simplify test, and exclude macOS (darwin)

thaJeztah:
Excluding darwin, as this package doesn't support it, but a developer may try running the tests on a Mac.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,13 +6,12 @@ all: lint test
 
 .PHONY: test
 test:
-	for p in $(PACKAGES); do \
-		(cd $$p && go test -v .); \
-	done
+	go test -v ./...
 
 .PHONY: lint
 lint: $(BINDIR)/golangci-lint
 	$(BINDIR)/golangci-lint version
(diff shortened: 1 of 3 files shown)


kolyshkin:
@thaJeztah have you checked if CI failures are related?

I'm afraid they are. This is a thing with two dependent packages from the same repo. If you want to test them together, you need to add `replace github.com/moby/sys/mountinfo ../mountinfo` in mount's `go.mod` for the test. I think it makes more sense to test packages separately, but I am open to discussion.

I suggest drop the makefile changes, but keep the package changes.

Or, add the above `replace` in the test.

thaJeztah:
Oh, thanks for the ping; yes, I think the failure is related, I realised this doesn't work after opening (but didn't dig deep yet).

Let me have a look at splitting that change

thaJeztah:
@kolyshkin opened https://github.com/moby/sys/pull/13

closing this one

PR #13, opened 2020-05-13
Exclude macOS (darwin)

thaJeztah:
Excluding darwin, as these packages don't support it, but a developer may try running the tests on a Mac.


thaJeztah:
@cpuguy83 ptal

PR #14, opened 2020-05-30
mountinfo: SingleEntryFilter support symlink

lujinda:
SingleEntryFilter supports matching cases with symlink in mountpoint.

In Linux, if there is a symlink in the mountpoint, the symlink will be resolved in the mountinfo file. mountpoint matching will fail.

For example:
```console
ln -s /mnt/ /tmp/mnt_symlink
mount -t ext4 /dev/vdb1 /tmp/mnt_symlink/disk1
```
What you see in mountinfo is `/mnt/disk1`, not `/tmp/mnt_symlink/disk1`.   
mountinfo.Mounted("/tmp/mnt_symlink/disk1") will return `false`, not `true`

### DIFF
--- a/mountinfo/mountinfo_filters.go
+++ b/mountinfo/mountinfo_filters.go
@@ -1,6 +1,9 @@
 package mountinfo
 
-import "strings"
+import (
+	"path/filepath"
+	"strings"
+)
 
 // FilterFunc is a type defining a callback function for GetMount(),
 // used to filter out mountinfo entries we're not interested in,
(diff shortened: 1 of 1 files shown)


thaJeztah:
@kolyshkin PTAL

mountinfo/mountinfo_filters.go
```diff
@@ -26,6 +29,11 @@ func PrefixFilter(prefix string) FilterFunc {
 // SingleEntryFilter looks for a specific entry
 func SingleEntryFilter(mp string) FilterFunc {
 	return func(m *Info) (bool, bool) {
+		// Resolve any symlinks in mountpoint, kernel would do the same and use the resolved path in /proc/<pid>/mountinfo.
+		if resolved, err := filepath.EvalSymlinks(mp); err == nil {
```
kolyshkin, line 33: This should probably be `filepath.Abs()`, which does it all.

And I think this is the responsibility of the caller to do that.

kolyshkin:
Maybe the better way would be to document that this package require absolute and cleaned/resolved paths. It would be a waste to perform an additional `filepath.Abs()` or `filepath.Clean()`.

mountinfo/mountinfo_filters.go
kolyshkin: I mixed it up, surely `filepath.Abs()` does not resolve symlinks, so you need `EvalSymlinks`, but maybe first you need to reject non-absolute paths.

And yes, I still think this is the responsibility of the caller.

lujinda: I agree that this is the caller’s responsibility。
Maybe we can execute `EvalSymlinks` in the `Mounted` function. Because Mounted returns whether a directory is a mountpoint.

PR #15, opened 2020-06-18
mountinfo: more docs

kolyshkin:
Closes: #14

### DIFF
--- a/mountinfo/doc.go
+++ b/mountinfo/doc.go
@@ -1,0 +1,48 @@
+// Package mountinfo provides a set of functions to retrieve information about OS mounts.
+// Currently it supports Linux. For historical reasons, there is also some support for FreeBSD,
+// and a shallow implementation for Windows, but in general this is Linux-only package, so
+// the rest of the document only applies to Linux, unless explicitly specified otherwise.
+//
+// In Linux, information about mounts seen by the current process is available from
+// /proc/self/mountinfo. Note that due to mount namespaces, different processes can
+// see different mounts. A per-process mountinfo table is available from /proc/<PID>/mountinfo,
+// where <PID> is a numerical process identifier.
+//
+// In general, /proc is not a very effective interface, and mountinfo is not an exception.
+// For example, there is no way to get information about a specific mount point (i.e. it
+// is all-or-nothing). This package tries to hide the /proc ineffectiveness by using
+// parse filters while reading mountinfo. A filter can skip some entries, or stop
+// processing the rest of the file once the needed information is found.
+//
+// For functions that have path as an argument (such as Mounted or various filters),
+// the argument must be
+//  - an absolute path;
+//  - having all symlinks resolved;
+//  - being cleaned.
+//
+// One way to achieve all of the above is to employ filepath.Abs followed by
+// filepath.EvalSymlinks (the latter calls filepath.Clean on the result so
+// there is no need to explicitly call filepath.Clean).
+//
+// NOTE that in many cases there is no need to consult mountinfo at all. Here are some
+// of the cases where mountinfo should not be parsed:
+//
+// 1. Before performing a mount. Usually, this is not needed, but if required (say to
+//    prevent overmounts), to check whether a directory is mounted, call os.Lstat
+//    on it and its parent directory, and compare their st.Sys().(*syscall.Stat_t).Dev
+//    fields -- if they differ, then the directory is the mount point. NOTE this does
+//    not work for bind mounts. Optionally, the filesystem type can also be checked
+//    by calling unix.Statfs and checking the Type field (i.e. filesystem type).
+//
+// 2. After performing a mount. If there is no error returned, the mount succeeded;
+//    checking the mount table for a new mount is redundant and expensive.
+//
+// 3. Before performing an unmount. It is more efficient to do an unmount and ignore
+//    a specific error (EINVAL) which tells the directory is not mounted.
+//
+// 4. After performing an unmount. If there is no error returned, the unmount succeeded.
+//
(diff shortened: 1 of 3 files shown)


kolyshkin:
@thaJeztah PTAL

kolyshkin:
@cpuguy83 PTAL

mountinfo/doc.go
```diff
@@ -0,0 +1,41 @@
...
+// see different mounts. A per-process mountinfo table is available from /proc/<PID>/mountinfo,
+// where <PID> is a numerical process identifier.
+//
+// In general, /proc is not a very effective interface, and mountinfo is not an exception.
+// For example, there is no way to get information about a specific mount point (i.e. it
+// is all-or-nothing). This package tries to hide the /proc ineffectiveness by using
+// parse filters while reading mountinfo. A filter can skip some entries, or stop
+// processing the rest of the file once the needed information is found.
+//
+// For any functions that have path as an argument (such as Mounted or various filters),
```
thaJeztah, line 17:
```suggestion
// For functions that have path as an argument (such as Mounted or various filters),
```

mountinfo/doc.go
```diff
@@ -0,0 +1,41 @@
...
+// the argument must be an absolute path, with all symlinks resolved, and cleaned.
```
thaJeztah, line 18:
```suggestion
// the argument must be an absolute path, with all symlinks resolved and cleaned.
```

mountinfo/doc.go
```diff
@@ -0,0 +1,41 @@
...
+// One way to achieve this is to process the path using filepath.EvalSymlinks.
+//
+// NOTE that in many cases there is no need to consult mountinfo at all. Here are some
+// of the cases where mountinfo should not be parsed:
+//
+// 1. Before performing a mount. If you want to check whether a directory is mounted,
+//    call os.Lstat on it and its parent directory, and compare their
+//    st.Sys().(*syscall.Stat_t).Dev fields -- if they differ, then the directory is
+//    the mount point. Optionally, you can also call unix.Statfs and check the Type
+//    field (i.e. filesystem type).
```
thaJeztah, line 28: Would it be useful to have utilities for this? Or too hard to have some thing that's universal enough?

mountinfo/doc.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// 2. After performing a mount. If there is no error returned, the mount succeded;
```
thaJeztah, line 30:
```suggestion
// 2. After performing a mount. If there is no error returned, the mount succeeded;
```

mountinfo/mountinfo.go
```diff
@@ -18,6 +18,10 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
 
 // Mounted determines if a specified mountpoint has been mounted.
 // On Linux it looks at /proc/self/mountinfo.
+//
+// The argument must be an absolute path, with all symlinks resolved, and clean.
```
thaJeztah, line 22:
```suggestion
// The argument must be an absolute path, with all symlinks resolved and cleaned.
```

mountinfo/mountinfo.go
```diff
@@ -18,6 +18,10 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
...
+// One way to ensure it is to process the path using filepath.EvalSymlinks before
```
thaJeztah, line 23: `it` is a bit ambiguous here, perhaps combine with the previous sentence?

`for example, using filepath.EvalSymlinks()`

mountinfo/doc.go
kolyshkin: the "clean" part is not about symlinks, but about whatever filepath.Clean() does. If we drop a comma here, it sounds like "cleaned" is about symlinks.

mountinfo/doc.go
kolyshkin: I actually plan to change Mounted() to use the above algo, similar to https://github.com/cilium/cilium/pull/12249.

Let me actually do this...

kolyshkin: https://github.com/moby/sys/pull/20

If we'll merge that one before this, I'll change the text here.

mountinfo/doc.go
kolyshkin: fixed

mountinfo/doc.go
kolyshkin: fixed

mountinfo/mountinfo.go
kolyshkin: I have worded it a bit differently

mountinfo/doc.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// 2. After performing a mount. If there is no error returned, the mount succeeded;
+//    checking the mount table for a new mount is redundant and expensive.
+//
+// 3. Before performing an unmount. It is more efficient to do an unmount and ignore
+//    a specific error (EINVAL) which tells the directory is not mounted.
+//
+// 4. After performing an unmount. If there is no error returned, the unmount succeeded.
+//
+// 5. To find the mount point root of a specific directory. You can perform os.Stat()
```
cpuguy83, line 44: Nice tip!

kolyshkin: Using this in runc; will probably add a function to do that in here.

mountinfo/doc.go
thaJeztah: Thanks!

mountinfo/doc.go
thaJeztah: Ah, yes, makes sense

Issue #17, opened 2020-06-23
Failed to run TestGetMounts in chroot


zhsj:
Background:

I packaged this library in Debian[1], and Ubuntu synced it to its archive[2]. But I found in Ubuntu it failed to build. After inspecting the log[3], I think the reason is `TestGetMounts`.

```
TestGetMounts: mountinfo_test.go:22: / should be mounted at least
```

Ubuntu builds this library with chroot. So I test it in a chroot and add some logs.

```
    TestGetMounts: mountinfo_test.go:15: &mountinfo.Info{ID:722, Parent:27, Major:0, Minor:5, Root:"/", Mountpoint:"/proc", Opts:"rw,relatime", Optional:"shared:177", Fstype:"proc", Source:"proc", VfsOpts:"rw"}
    TestGetMounts: mountinfo_test.go:15: &mountinfo.Info{ID:738, Parent:27, Major:0, Minor:20, Root:"/", Mountpoint:"/sys", Opts:"rw,nosuid,nodev,noexec,relatime", Optional:"shared:7", Fstype:"sysfs", Source:"sysfs", VfsOpts:"rw"}
    TestGetMounts: mountinfo_test.go:15: &mountinfo.Info{ID:739, Parent:738, Major:0, Minor:7, Root:"/", Mountpoint:"/sys/kernel/security", Opts:"rw,nosuid,nodev,noexec,relatime", Optional:"shared:8", Fstype:"securityfs", Source:"securityfs", VfsOpts:"rw"}
    TestGetMounts: mountinfo_test.go:15: &mountinfo.Info{ID:740, Parent:738, Major:0, Minor:25, Root:"/", Mountpoint:"/sys/fs/cgroup", Opts:"ro,nosuid,nodev,noexec", Optional:"shared:9", Fstype:"tmpfs", Source:"tmpfs", VfsOpts:"ro,mode=755"}
    TestGetMounts: mountinfo_test.go:15: &mountinfo.Info{ID:741, Parent:740, Major:0, Minor:26, Root:"/", Mountpoint:"/sys/fs/cgroup/unified", Opts:"rw,nosuid,nodev,noexec,relatime", Optional:"shared:10", Fstype:"cgroup2", Source:"cgroup2", VfsOpts:"rw,nsdelegate"}
[...]

PR #18, opened 2020-06-24
mountinfo/TestGetMounts: fix for Ubuntu build system

kolyshkin:
Apparently, Ubuntu build system is some sort of a chroot which does not
have `/` mount in its /proc/self/mountinfo. This leads to test failure
described in https://github.com/moby/sys/issues/17.

The fix is not to require a specific mount (even `/` that must exist but
apparently does not in some setups). Instead, let's require at least 2
mounts, doesn't matter which ones.

For reference, his test case was added by
https://github.com/moby/moby/pull/4812

Fixes:  https://github.com/moby/sys/issues/17

### DIFF
--- a/mountinfo/mountinfo_test.go
+++ b/mountinfo/mountinfo_test.go
@@ -10,14 +10,7 @@ func TestGetMounts(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	root := false
-	for _, entry := range mounts {
-		if entry.Mountpoint == "/" {
-			root = true
-		}
-	}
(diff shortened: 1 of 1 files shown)


kolyshkin:
@thaJeztah @cpuguy83 PTAL

kolyshkin:
The Linux kernel's show_mountinfo() has a check that the mount is accessible, and it is not shown if not. Looks like this is why the mount for / is not shown in the Ubuntu case -- they have some weird setup.

PR #19, opened 2020-06-25
Fix failing test and the false negative CI

zhsj:


mountinfo/mountinfo_linux.go
```diff
@@ -194,7 +194,7 @@ func unescape(path string) (string, error) {
 			v := c - '0'
 			for j := 2; j < 4; j++ { // one digit already; two more
 				x := s[j] - '0'
-				if x < 0 || x > 7 {
```
zhsj, line 197: It's changed because

```
mountinfo_linux.go:197:8: SA4003: no value of type byte is less than 0 (staticcheck)
```

SamWhited:
Would setting `.POSIX:` also fix this? I normally stick this at the top of my makefiles since it's portable between GNU Make and NetBSD Make (though that doesn't really matter here).

zhsj:
> Would setting `.POSIX:` also fix this? I normally stick this at the top of my makefiles since it's portable between GNU Make and NetBSD Make (though that doesn't really matter here).

I just know this `.POSIX:` syntax.. However IMO it's less clear than `.SHELLFLAGS`.

mountinfo/mountinfo_linux.go
```diff
@@ -194,7 +194,7 @@ func unescape(path string) (string, error) {
...
+				if x > 7 {
```
kolyshkin, line 197: OK I think the correct thing to do would be
```go
if s[j] < '0' || s[j] > '7' {
  return "", fmt.Errorf("bad escape sequence %q: not a digit", s[:3])
}
x := s[j] - '0'
```

kolyshkin:
Thank you for the fixes! Can you please split it into 3 commits:

- Fixing the test case
- Fixing the is-octal-digit check
- fixing the make flags

with a brief commit message describing the reason for the change.

Otherwise it won't be easy to follow the git history

mountinfo/mountinfo_linux.go
kolyshkin: This won't fix anything (we will still catch underflow in case s[j] is less than '0'), but it's a tad cleaner code with no ambiguity.

mountinfo/mountinfo_linux.go
kolyshkin: Do you mind creating a PR adding this linter (or linters) to this repo CI?

kolyshkin: Oh OK nevermind, it's there already but the non-zero exit code is currently ignored by make.

zhsj:
> Thank you for the fixes! Can you please split it into 3 commits:
> 
> * Fixing the test case
> * Fixing the is-octal-digit check
> * fixing the make flags
> 
> with a brief commit message describing the reason for the change.
> 
> Otherwise it won't be easy to follow the git history

done

PR #20, opened 2020-06-29
mountinfo.Mounted: optimize by adding fast paths

kolyshkin:
Add two fast paths to avoid (slow) mountinfo parsing:

1. In case path does not exist, it is definitely NOT a mount point.

2. In case path's `st.Dev` differs from that of its parent, it is
   definitely a mount point.

In all the other cases (including bind mounts, case of `path="/"`, other
errors from `unix.Lstat`) the code falls back to the old slow method
of parsing mountinfo.

### DIFF
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -16,19 +16,13 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseInfoFile(reader, f)
 }
 
-// Mounted determines if a specified mountpoint has been mounted.
-// On Linux it looks at /proc/self/mountinfo.
+// Mounted determines if a specified path is a mount point.
 //
 // The argument must be an absolute path, with all symlinks resolved, and clean.
-// One way to ensure it is to process the path using filepath.EvalSymlinks before
-// calling this function.
-func Mounted(mountpoint string) (bool, error) {
-	entries, err := GetMounts(SingleEntryFilter(mountpoint))
-	if err != nil {
-		return false, err
-	}
-
-	return len(entries) > 0, nil
+// One way to ensure it is to process the path using filepath.Abs followed by
+// filepath.EvalSymlinks before calling this function.
+func Mounted(path string) (bool, error) {
+	return mounted(path)
 }
 
 // Info reveals information about a particular mounted filesystem. This
--- a/mountinfo/mountinfo_unix.go
+++ b/mountinfo/mountinfo_unix.go
@@ -1,0 +1,35 @@
+// +build linux freebsd,cgo
+
+package mountinfo
+
+import (
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+func mounted(path string) (bool, error) {
+	var st unix.Stat_t
+
+	err := unix.Lstat(path, &st)
+	switch err {
+	case unix.ENOENT:
+		// Nonexistent path, so not a mount point.
+		return false, nil
+	case nil:
+		dev := st.Dev
+		err = unix.Lstat(filepath.Dir(path), &st)
+		if err == nil && dev != st.Dev {
+			// Device number differs from that of parent,
+			// so definitely a mount point.
+			return true, nil
+		}
+	}
+
+	entries, err := GetMounts(SingleEntryFilter(path))
+	if err != nil {
+		return false, err
+	}
+
+	return len(entries) > 0, nil
+}
--- a/mountinfo/go.mod
+++ b/mountinfo/go.mod
@@ -1,3 +1,5 @@
 module github.com/moby/sys/mountinfo
 
 go 1.14
+
+require golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae
--- a/mountinfo/go.sum
+++ b/mountinfo/go.sum
@@ -1,0 +1,2 @@
+golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae h1:/WDfKMnPU+m5M4xB+6x4kaepxRw6jWvR5iDRdvjHgy8=
+golang.org/x/sys v0.0.0-20200223170610-d5e6a3e2c0ae/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
--- a/mountinfo/mountinfo_unsupported.go
+++ b/mountinfo/mountinfo_unsupported.go
@@ -8,10 +8,16 @@ import (
 	"runtime"
 )
 
+var errNotImplemented = fmt.Errorf("not implemented on %s/%s", runtime.GOOS, runtime.GOARCH)
+
 func parseMountTable(_ FilterFunc) ([]*Info, error) {
-	return nil, fmt.Errorf("mount.parseMountTable is not implemented on %s/%s", runtime.GOOS, runtime.GOARCH)
+	return nil, errNotImplemented
 }
 
 func parseInfoFile(_ io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseMountTable(f)
+}
+
+func mounted(path string) (bool, error) {
+	return false, errNotImplemented
 }
--- a/mountinfo/mountinfo_windows.go
+++ b/mountinfo/mountinfo_windows.go
@@ -10,3 +10,7 @@ func parseMountTable(_ FilterFunc) ([]*Info, error) {
 func parseInfoFile(_ io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseMountTable(f)
 }
+
+func mounted(_ string) (bool, error) {
+	return false, nil
+}


mountinfo/mountinfo.go
```diff
@@ -16,15 +22,31 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseInfoFile(reader, f)
 }
 
-// Mounted determines if a specified mountpoint has been mounted.
-// On Linux it looks at /proc/self/mountinfo.
-func Mounted(mountpoint string) (bool, error) {
-	entries, err := GetMounts(SingleEntryFilter(mountpoint))
+// Mounted determines if a directory specified by path has been mounted.
+func Mounted(path string) (bool, error) {
+	var st, pst unix.Stat_t
```
cpuguy83, line 27: Looks like we need to split this for the windows case.

mountinfo/mountinfo.go
```diff
@@ -16,15 +22,31 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseInfoFile(reader, f)
 }
 
-// Mounted determines if a specified mountpoint has been mounted.
-// On Linux it looks at /proc/self/mountinfo.
-func Mounted(mountpoint string) (bool, error) {
-	entries, err := GetMounts(SingleEntryFilter(mountpoint))
+// Mounted determines if a directory specified by path has been mounted.
```
tianon, line 25: The old/current version of this function returns appropriate answers for mounted files as well as directories, right?  Are any of the existing usages of it sensitive to that distinction / would be broken by this change?

kolyshkin: Whoops, I entirely forgot about bind-mounted files scenario. Surely, this check needs to be removed.

mountinfo/mountinfo.go
kolyshkin: Right; will do

kolyshkin:
So, the method is not working for bind mounts, so I changed the commit to try comparing devices first, then falling back to mountinfo parsing. This way, if there is a proper (not bind) mount, we can skip parsing mountinfo.

mountinfo/mountinfo_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+// Mounted determines if path is mounted (i.e. it is a mount point).
+func Mounted(path string) (bool, error) {
+	// Try the fast method (of comparing devices) first.
+	// Alas, it can't detect bind mounts, so fall back on
+	// good ol' mountinfo parsing.
```
thaJeztah, line 15: Not sure this is the best location for this comment; thinking if it should either be;

- As part of the GoDoc comment (describing the strategy)
- Further down (before `entries, err := GetMounts(SingleEntryFilter(path))`), describing that that's where we fail back to parsing MountInfo (which may be needed if the path is a bind-mount)

mountinfo/mountinfo_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	var st, pst unix.Stat_t
+
+	err := unix.Lstat(path, &st)
```
thaJeztah, line 18: Just noticed; `unix.Lstat` is not available on macOS. While this function was never functional on macOS, it looks like it did _compile_ before (but called out to `parseMountTable()`,  which returned an error)
https://github.com/moby/sys/blob/0783079dd297647b492eef4d729d657cc7d3bc6c/mountinfo/mountinfo_unsupported.go#L11-L13

Same for freebsd https://github.com/moby/sys/blob/0783079dd297647b492eef4d729d657cc7d3bc6c/mountinfo/mountinfo_unsupported.go#L1

mountinfo/mountinfo_unix.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return false, err
+	}
+	err = unix.Lstat(filepath.Dir(path), &pst)
+	if err != nil {
```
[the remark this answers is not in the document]
kolyshkin: The code that calls this usually does mount/unmount next, and for that, root is required.

And yet I was thinking that maybe instead of returning an error we can just jump to the fallback code of parsing mountinfo.

mountinfo/mountinfo_unix.go
kolyshkin: right

mountinfo/mountinfo_unix.go
kolyshkin: AFAIK Darwin does have lstat(2), for example, see

https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/stat.2.html

Have you tried it?

mountinfo/mountinfo.go
```diff
@@ -16,17 +16,6 @@ func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
 	return parseInfoFile(reader, f)
 }
 
-// Mounted determines if a specified mountpoint has been mounted.
-// On Linux it looks at /proc/self/mountinfo.
-func Mounted(mountpoint string) (bool, error) {
```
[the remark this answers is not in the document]
kolyshkin: done (for Mounted)

kolyshkin:
I have reworked it considerably, PTAL @tianon @thaJeztah

mountinfo/mountinfo_unix.go
thaJeztah: Arf, looks like my go modules cache was probably out of date 😞

I do see that the build tags need updating in your latest version; for `darwin`, `mounted()` is now "redeclared"

thaJeztah: the _unsupported.go has `// +build !windows,!linux,!freebsd freebsd,!cgo`, so should probably have `!darwin` added, but doing so would then make `parseMountTable` and `parseInfoFile` go missing on `darwin` (because they're in the `_linux.go` file 😞

thaJeztah:
Looks like there's still a duplicate declaration for `mounted()` on darwin;

file                  | darwin             | freebsd            | linux              | windows
----------------------|--------------------|--------------------|--------------------|--------------------
mountinfo_unix        | :white_check_mark: | :white_check_mark: | :white_check_mark: | -
mountinfo_unsupported | :white_check_mark: | -                  | -                  | -
mountinfo_windows     | -                  | -                  | -                  | :white_check_mark:
conflict?             | :warning:          | :white_check_mark: | :white_check_mark: | :white_check_mark:

mountinfo/mountinfo_unix.go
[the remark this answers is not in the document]
kolyshkin: In fact you were right. If path does not exist, this means it is not mounted (since a mount point can not be removed).

In all the other cases I fall back to mountinfo parsing

thaJeztah:
Cross is failing; 

```
for os in linux freebsd darwin windows; do \
	for arch in amd64 arm arm64 ppc64le s390x; do \
		echo "$os/$arch" | grep -qE '^((freebsd|darwin|windows)/(ppc64le|s390x)|windows/arm64|freebsd/s390x)$' && continue; \
		echo "# building for $os/$arch"; \
		GOOS=$os GOARCH=$arch go build ./...; \
	done; \
done
# building for linux/amd64
##[error]mount/mount.go:9:2: cannot find package "github.com/moby/sys/mountinfo" in any of:
	/opt/hostedtoolcache/go/1.13.12/x64/src/github.com/moby/sys/mountinfo (from $GOROOT)
	/home/runner/go/src/github.com/moby/sys/mountinfo (from $GOPATH)
##[error]mount/flags_linux.go:4:2: cannot find package "golang.org/x/sys/unix" in any of:
	/opt/hostedtoolcache/go/1.13.12/x64/src/golang.org/x/sys/unix (from $GOROOT)
	/home/runner/go/src/golang.org/x/sys/unix (from $GOPATH)
##[error]Makefile:33: recipe for target 'cross' failed
make: *** [cross] Error 1
##[error]Process completed with exit code 2.
```

Does it need the same trickery as we use for `test`? https://github.com/moby/sys/blob/fb9f8cf904285111be4c0db77af0ee5b3b4fdba3/Makefile#L10-L12


Starting to wonder if it would be cleaner to have separate Makefiles in each module, and then call those targets from the main Makefile (also beginning to feel the pain of the submodules 😓 - do you think it causes more issues to have them in the same repository than to have to maintain two repositories?)

kolyshkin:
Will fix

> Starting to wonder if it would be cleaner to have separate Makefiles in each module

It will result in more code (same/similar `for` in the top makefile, plus `go test` in every sub-makefile).

kolyshkin:
OK we need `replace` in go.mod for such tests, and once `replace` is there, we can ditch the `for p in $(PACKAGES)` in the top Makefile.

Makefile
```diff
@@ -24,3 +27,15 @@ $(BINDIR)/golangci-lint: $(BINDIR)
 
 $(BINDIR):
 	mkdir -p $(BINDIR)
+
+.PHONY: cross
+cross:
+	for os in $(CROSS_OSES); do \
+		for arch in $(CROSS_ARCHES); do \
+			echo "$$os/$$arch" | grep -qE $(OS_ARCH_SKIP) && continue; \
```
thaJeztah, line 35: Looks like the list is;

```
linux/amd64
linux/arm
linux/arm64
linux/ppc64le
linux/s390
freebsd/amd64
freebsd/arm
darwin/amd64
darwin/arm
darwin/arm64
windows/amd64
windows/arm
```

Wondering if we need:

- freebsd/arm
- darwin/arm (likely not, darwin won't be 32-bit)
- windows/arm (not sure; is windows on 32-bit arm a thing, or are they working on 64 bit only?)

If we remove those from the list;

```
linux/amd64
linux/arm
linux/arm64
linux/ppc64le
linux/s390
freebsd/amd64
darwin/amd64
darwin/arm64
windows/amd64
```

Wondering if it would be better to just define that list (easier read, and easier to maintain as well)

If would require splitting to GOOS and GOARCH then, so something like

```bash
GOARCH=${CROSS#*/}
GOOS=${CROSS%/*}
```

Or handle it with Make;

```makefile
.PHONY: linux/% darwin/% freebsd/% windows/%
linux/% darwin/% freebsd/% windows/%:
        echo GOOS=$(@D)
        echo GOARCH=$(@F)
```

kolyshkin: Indeed this looks cleaner. Done; PTAL

kolyshkin:
Separated out cross build to https://github.com/moby/sys/pull/23

thaJeztah:
@cpuguy83 PTAL

thaJeztah:
oh, and @tianon 🤗

mountinfo/mountinfo_unix.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func mounted(path string) (bool, error) {
+	var st unix.Stat_t
+
+	err := unix.Lstat(path, &st)
+	switch err {
+	case unix.ENOENT:
+		// Nonexistent path, so not a mount point.
+		return false, nil
```
tianon, line 18: I had some concern that a "broken" FUSE mount might hit this case, but it returns `ENOTCONN`. :+1:

(Just including a note in case anyone else reading this has the same question/concern. :sweat_smile:)

kolyshkin:
@cpuguy83 PTAL

Issue #21, opened 2020-06-30
is RecursiveUnmount needed?


kolyshkin:
Today I was looking into RecursiveUnmount code and thought we can optimize it by adding the fast path, like this:

```diff
 // RecursiveUnmount unmounts the target and all mounts underneath, starting with
 // the deepsest mount first.
 func RecursiveUnmount(target string) error {
+       // Fast path: if we can unmount target, there are no submounts!
+       if err := unmount(target, 0); err == nil {
+               return nil
+       }
+
        mounts, err := mountinfo.GetMounts(mountinfo.PrefixFilter(target))
        if err != nil {
                return err
```

The idea is, if there are other mounts under `target`, we'll go the usual slow way, but if not, all it takes a single syscall.

When I read the umount(2) man page for the umphteen time, to confirm my gut feeling I should NOT use MNT_DETACH in the fast path, and found this gem:

> immediately disconnect the filesystem and  all  filesystems
> mounted below it from each other and from the mount table

This means that MNT_DETACH performs a recursive unmount. Since we do use MNT_DETACH from RecursiveUnmount anyway, it seems all of it, in its entirety, can be replaced by a single `umount(2)` with MNT_DETACH.

@cpuguy83 can you please shed some light on how you came with RecursiveUnmount in the first place? I tracked its inception down to PR  https://github.com/moby/moby/pull/31012 (https://github.com/moby/moby/pull/31012/commits/54dcbab25ea4771da303fa95e0c26f2d39487b49#diff-ce9e7d50400d88435b2be4f4cd9b7b23R61) but it's not very clear why it was added. I am surely very interested in your other input on this topic.

cpuguy83:
The problem with `MNT_DETACH` is we can end up leaking data if there are leaked mounts... e.g. we try to remove a directory we did a lazy unmount of, it succeeds, but the reference is still held in another namespace so the data isn't actually gone.

The primary case we found for this issue was other system services, such as ntpd, get started with their own  mount ns and wind up holding onto mount references which docker made private and therefore unable to be unmounted fully, and as such the data never actually cleaned up.

As to why was `RecursiveUnount` created, indeed to battle some of this which was not fully understood at the time. Additionally there can technically be mounts that a container creates that we don't know about which get propagated to the host ns.

kolyshkin:
Yep, I'm well aware of `MNT_DETACH` limitations, as well as the issues wrt making mounts that leaked to other mount namespaces private.

I guess, my ultimate question is, since

[...]

PR #26, opened 2020-08-27
mount.RecursiveUnmount: add a fast path

kolyshkin:
In many cases, this is called with an argument that is a mount point
itself. In such case, if we're on Linux, MNT_DETACH ensures that the
umount is recursive (which is its nice, but a little known feature).
For other platforms, if the mount does not have any submounts
underneath and is not busy, it will be unmounted.
    
In the above case, we no longer have to parse mountinfo and perform
the unmounts one-by-one.
    
IOW this should speed up RecursiveUnmount() for some cases, without
breaking the compatibility.
    
While at it
 - fix a typo in doc;
 - add that the argument does not have to be a mount point;
 - improve mount.Unmount doc.

Fixes: https://github.com/moby/sys/issues/21

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -17,8 +17,8 @@ func Mount(device, target, mType, options string) error {
 	return mount(device, target, mType, uintptr(flag), data)
 }
 
-// Unmount lazily unmounts a filesystem on supported platforms, otherwise
-// does a normal unmount.
+// Unmount lazily unmounts a filesystem on supported platforms, otherwise does
+// a normal unmount.  If target is not a mount point, no error is returned.
 func Unmount(target string) error {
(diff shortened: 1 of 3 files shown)


kolyshkin:
@cpuguy83 PTAL

AkihiroSuda:
Will this introduce a risk of having orphan mounts on error?

kolyshkin:
> Will this introduce a risk of having orphan mounts on error?

If your concern is about MNT_DETACH, we use this flag anyway, so AFAIU we're not making anything worse. Please see a longer discussion in #21.

If it is something different, please explain your concern in more details.

AkihiroSuda (approved):

Issue #27, opened 2020-08-27
Mounted: add path resolution


kolyshkin:
The argument to `mount.Mounted` call needs to be an absolute path with symlinks resolved and cleaned (otherwise it won't match the path in mountinfo). Currently, it is a caller's duty to make sure the path is prepared. I am thinking into moving this functionality into the library, because
1. if the path is already prepared, this won't affect performance much (it's a few `stat(2)` syscalls);
2. if the path is not prepared and is resolved correctly, we avoid "false negative".

The implementation of 'resolve' is something like this:

```go
        if realPath, err = filepath.Abs(path); err != nil {
                return false, fmt.Errorf("unable to get absolute path for %q: %w", path, err)
        }
        if realPath, err = filepath.EvalSymlinks(realPath); err != nil {
                return false, fmt.Errorf("failed to canonicalise path for %q: %w", path, err)
        }
        if _, err := os.Stat(realPath); err != nil {
                return false, fmt.Errorf(err, "failed to stat target %q of %q: %w", realPath, path, err)
        }
```

kolyshkin:
Well, item 2 above is not applicable as we already `Stat` the target, so we require it to exist and be resolvable.

kolyshkin:
Also, we can't do the same "resolve" for mount filters as there are no way to return an error from a filter, so the caller is still responsible.

Issue #25, opened 2020-08-24
Implement Mounted() optimization using openat2(RESOLVE_NO_XDEV)


kolyshkin:
Currently, `Mounted()` have to consult mountinfo to figure out if the argument is a mount point.

This mountinfo parsing can be avoided by doing two stat() calls and comparing `dev_t`. Unfortunately, it is not working for bind mounts.

Now, with `openat2(RESOLVE_NO_XDEV)`, we might be able to implement a trick to do a similar check for bind mounts, thus eliminating the need to parse mountinfo entirely.

cyphar:
You can use `RESOLVE_NO_XDEV` for both bind-mounts and regular mountpoints -- `RESOLVE_NO_XDEV` disables any `vfsmount` jumps -- which is what you want to detect here. Plus it's only one syscall to check each mountpoint (though ideally you'd do the check by first opening the directory then doing `openat2(dirfd, "..", {.resolve = RESOLVE_NO_XDEV})` -- since if you used whole paths you'd get `-EXDEV` if *any* component is a mountpoint crossing)

kolyshkin:
Ah, I totally forgot I tried adding openat2 support to x/sys/unix and it got reverted.

Trying one more time: https://go-review.googlesource.com/c/sys/+/253057

(surely I can just use `unix.Syscall()` directly and define a few constants, but I'd rather not)

kolyshkin:
I guess we have to check first if `path` is a file or directory. For non-directories, we need to use file's directory as dirfd and the file itself as a file. For directories, it is as described above

kolyshkin:
Funny, it requires 5 syscalls.

1. `lstat` to figure out whether it's a directory or not.
2. `open` to open a directory.
3. `openat2` -- the actual check.
4. close for dirfd obtained in 2.
5. close for fd obtained in 3.

PR #28, opened 2020-08-27
mount.Mounted: normalize the path argument

kolyshkin:
Currently, the path provided to mount.Mounted() must be
 - an absolute path;
 - with all symlinks resolved;
 - cleaned.

Otherwise, the function will most probably return a false negative in
case of a bind mount (i.e. "not mounted" while the path is in fact
mounted).

This commit removes the above limitation by doing a path normalization.
This should slow things down but it's not affecting the fast path, so
the slowdown will only be seen in the bind mount case.

Fixes: https://github.com/moby/sys/issues/27

### DIFF
--- a/mountinfo/mountinfo_unix.go
+++ b/mountinfo/mountinfo_unix.go
@@ -3,10 +3,25 @@
 package mountinfo
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"golang.org/x/sys/unix"
 )
+
+func normalizePath(path string) (realPath string, err error) {
+	if realPath, err = filepath.Abs(path); err != nil {
+		return "", fmt.Errorf("unable to get absolute path for %q: %w", path, err)
+	}
+	if realPath, err = filepath.EvalSymlinks(realPath); err != nil {
+		return "", fmt.Errorf("failed to canonicalise path for %q: %w", path, err)
+	}
+	if _, err := os.Stat(realPath); err != nil {
+		return "", fmt.Errorf("failed to stat target %q of %q: %w", realPath, path, err)
+	}
+	return realPath, nil
+}
 
 func mounted(path string) (bool, error) {
 	var st unix.Stat_t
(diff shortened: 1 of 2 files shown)


kolyshkin:
NOTE I'm not sure we should merge this, just proposing for a discussion

kolyshkin:
@cpuguy83 @thaJeztah PTAL

mountinfo/mountinfo_unix.go
```diff
@@ -26,6 +41,11 @@ func mounted(path string) (bool, error) {
 		}
 	}
 
+	path, err = normalizePath(path)
```
thaJeztah, line 44: A bit on the fence if we should do the magic as part of this, or instead provide "normalizepath" as a utility, so that the caller can decide wether or not paths should be resolved

kolyshkin: Good idea about providing `NormalizePath()` -- perhaps this is the way to go.

Will implement

kolyshkin: In some cases though we don't need a normalized path (for example, when this is not a bind mount, a couple of stat() calls is sufficient and there's no need to perform "path normalization". There will be another optimization (see https://github.com/moby/sys/issues/25) that, also, does not require path normalization.

So if we ask a user to call it (in case they are not sure if the path is "normalized"), and one of the above optimization is used, the call is useless, thereas if we call it in the middle of `mounted()` like this patch does, the call is not useless.

I dunno :dog:

thaJeztah: Hmm.. right; should it be implemented as a filter/functional parameter that can be passed somewhere? (thinking out loud)

kolyshkin: What I think now is
1. with the optimizations in #25 we will hit mountinfo parser rarely;
2. parsing mountinfo is much heavier (in terms of time/resources) than normalizing one path, so I guess if we add it there, the effect is negligible.

Let me rebase this one on top of #25

kolyshkin: Sorry I mean #29 not #25 of course

kolyshkin: Ideally, this functionality can be enabled by a mere binary flag (`NormalizeMePlease`), but we don't have `flags` argument and I don't want to add one.

kolyshkin: Decided to close this one and add the commit to #29 instead.

PR #29, opened 2020-09-08
mountinfo.Mounted: add fast path for Linux using openat2

kolyshkin:
A new (as of Linux 5.6) `openat2(2)` system call can be used to check
whether a path is a mount point or not, and it even works for bind
mounts!

This relies on the fact that `openat2()` with `RESOLVE_NO_XDEV` option
won't cross the filesystem boundary (that includes bind mounts).
For directories, we try to open a parent directory (`..`), and for
files, a directory the file is in.

The error from openat2 is handled in the following way:

* `EXDEV`:  the argument is definitely a mount point.
* no error: the argument is definitely not a mount point.

For all the other errors (most probably `ENOSYS` or `EPERM`), fall
back to existing methods (`mountedByStat`,  `mountedByMountinfo`).

Add a couple of tests:

* `TestMountedBy` prepares a few mounts, and checks that various
  `mountedBy*` functions work as expected. Requires root.

* `TestMountedByOpenat2VsMountinfo` reads the mountinfo and then tries
  to use `mountedByOpenat2()` for each mount, ignoring `EPERM` as the test
  might not be run as root. AFAIK this is the only way to test the
  new functionality if one is not root.

v2 changes:
 - drop our own openat2 wrapper;
 - add O_PATH to openat2 flags;
 - add O_CLOEXEC to all open* calls;
 - rm socket and symlink special cases;
 - rm some debug from tests.

v3 changes:
 - add normalization of path to `mountedByMountinfo` (moved from #28).

v4 changes:
 - get rid of dir/file distinction in `mountedByOpenat2` to simplify the code;
 - remove the second argument from `mountedByStat` and `mountedByOpenat2`;
 - treat ENOENT as "not mounted" by all `mountedBy*` functions

Fixes: https://github.com/moby/sys/issues/25
Fixes: https://github.com/moby/sys/issues/27


kolyshkin:
@cpuguy83 @AkihiroSuda @thaJeztah @cyphar PTAL

mountinfo/mounted_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	var dir, file string
+
+	ftype := st.Mode & unix.S_IFMT
+	switch ftype {
+	case unix.S_IFDIR:
+		dir = path
+		file = ".."
+	default:
+		dir, file = filepath.Split(path)
+	}
```
cyphar, line 22: If you need to open the directory separately anyway, you can implement both directory and regular file handling as `openat2(dfd, "file", ...)` -- though I guess it's the same either way.

thaJeztah:
Looks like https://go-review.googlesource.com/c/sys/+/253057/ was merged (https://github.com/golang/sys/commit/eff7692f900947b7d782d16af70ca32cc40774f0) (thanks @tklauser!), so this PR can remove the temporary wrapper

mountinfo/mounted_linux.go
```diff
@@ -0,0 +1,82 @@
...
+
+	dirfd, err := unix.Open(dir, unix.O_RDONLY|unix.O_PATH, 0)
```
cyphar, line 24: `O_RDONLY` isn't needed -- also probably should be `O_CLOEXEC` as well?

mountinfo/mounted_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return false, &os.PathError{Op: "open", Path: dir, Err: err}
+	}
+	fd, err := openat2(dirfd, file, &openHow{
+		Flags:   unix.O_RDONLY | unix.O_NOFOLLOW,
```
cyphar, line 29: This should also be `O_PATH` not `O_RDONLY`.

mountinfo/openat2_linux.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	RESOLVE_BENEATH       = 0x8
+	RESOLVE_IN_ROOT       = 0x10
+	RESOLVE_NO_MAGICLINKS = 0x2
+	RESOLVE_NO_SYMLINKS   = 0x4
+	RESOLVE_NO_XDEV       = 0x1
+)
+
+// FIXME: sycall numbers are arch-specific, this won't work on e.g. MIPS
+const SYS_OPENAT2 = 437
```
cyphar, line 26: Actually, `openat2` should have the same syscall number on all architectures (except maybe Alpha). There was a very recently push to make all newer syscalls have the same syscall numbers.

mountinfo/openat2_linux.go
```diff
@@ -0,0 +1,44 @@
...
+	"golang.org/x/sys/unix"
+)
+
+type openHow struct {
+	Flags   uint64
+	Mode    uint64
+	Resolve uint64
+}
+
+const sizeofOpenHow = 0x18
```
cyphar, line 15: A minor thing, but this makes me a little nervous -- is it possible to use `unsafe.Sizeof`?

mountinfo/openat2_linux.go
kolyshkin: This is temp POC code, I'll remove it now once https://go-review.googlesource.com/c/sys/+/253057/ is merged

mountinfo/openat2_linux.go
kolyshkin: 1. This is temp POC code.
2. This is generated code, the generator itself (in x/sys/unix) is actually using C sizeof.

mountinfo/mounted_linux.go
kolyshkin: Awesome, that'll remove these not-so-good-looking special cases.

mountinfo/mounted_linux.go
kolyshkin: I think that is what I'm doing (or I don't get what you mean here).

mountinfo/mounted_linux.go
kolyshkin: fixed

kolyshkin:
v2 changes (thanks to @cyphar for review):
 - drop our own openat2 wrapper;
 - add O_PATH to openat2 flags;
 - add O_CLOEXEC to all open* calls;
 - rm socket and symlink special cases;
 - rm some debug from tests.

kolyshkin:
@cpuguy83 @thaJeztah PTAL. I am mostly interested in any test cases I might be missing.

kolyshkin:
v3: add path normalization to mountedByMountinfo

mountinfo/mounted_linux.go
cyphar: I meant you could implement both cases (file and directory target) with `dir, file := filepath.Split(path)`. The `dir, file := path, ".."` trick I mentioned originally is an optimisation you can do if you already have the directory open, but if you don't have it open then both methods are equivalent.

But this is more of a nit than anything else, both ways work.

mountinfo/mounted_linux.go
```diff
@@ -0,0 +1,73 @@
...
+
+	switch st.Mode & unix.S_IFMT {
+	case unix.S_IFDIR:
+		dir = path
+		file = ".."
+	default:
+		dir, file = filepath.Split(path)
+	}
+
+	dirfd, err := unix.Open(dir, unix.O_PATH|unix.O_CLOEXEC, 0)
```
cyphar, line 23: This is more of a question than a requested change, but you care about symlinks in the path? If you don't want any symlinks in the path you could open `dirfd` with `openat2(RESOLVE_NO_SYMLINKS)` -- but I'm not sure what semantics you want for this function.

cyphar:
Aside from the two nits, LGTM.

mountinfo/mounted_linux.go
kolyshkin: I would like it to work with any paths, so let this stay as is.

And yet it makes sense to use `openat2()` here since that way we'll bail out earlier if it's not supported.

cpuguy83 (approved):
LGTM

@cyphar ❤️  for `openat2`. Love it.

cpuguy83:
Downside is I don't think we can test this in CI without a custom VM.

kolyshkin:
> Downside is I don't think we can test this in CI without a custom VM.

Should be relatively easy to do, @AkihiroSuda did it for runc using vagrant on Travis CI. I might take a look at it later.

mountinfo/mounted_linux.go
kolyshkin: That's right, not sure how I missed it earlier. Will update.

kolyshkin:
v4 changes:

- get rid of dir/file distinction in `mountedByOpenat2` to simplify the code;
- remove the second argument from `mountedByStat` and `mountedByOpenat2`;
- treat `ENOENT` as "not mounted" by all `mountedBy*` functions

mountinfo/mounted_linux.go
kolyshkin: Updated, simplified the code a lot!

kolyshkin:
> I might take a look at it later

Added :tada:

kolyshkin:
aaand it's a pass! https://travis-ci.org/github/moby/sys/builds/727240869#L4099

mountinfo/mounted_linux_test.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		err = &os.PathError{Op: "mount", Path: mnt, Err: err}
+		return
+	}
+	mounts = append(mounts, mnt)
+
+	return
+}
+
+func cleanupMounts(t *testing.T, dir string, mounts []string) {
```
thaJeztah, line 162: (not for this PR, but just thinking out loud); we should have a look at the `testing.Cleanup()` option that's now available in Go 1.13+

mountinfo/mounted_linux_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	dir, mounts, err := prepareMounts(t)
+	defer cleanupMounts(t, dir, mounts)
+	if err != nil {
+		t.Fatalf("setup failed: %v", err)
+	}
+
+	checked := false
+	for _, m := range mounts {
```
thaJeztah, line 189: (not a blocker, just thinking out loud as well) We could use subtests for these for easier debugging

thaJeztah (approved):
LGTM 👍

thaJeztah:
I think @cyphar's comments were addressed, so I'll go ahead and merge <3

Issue #24, opened 2020-07-02
drop freebsd support?


kolyshkin:
I propose to drop freebsd support from these packages (moby/sys/mount and moby/sys/mountinfo)

Pros:
 * it is not being tested and I'm not sure it's working
 * it is not used by moby/moby, and might not have any users at all
 * mountinfo for freebsd requires cgo (for [getmntinfo(3)](https://www.freebsd.org/cgi/man.cgi?query=getmntinfo))

Cons:
* it might still being used by someone

AkihiroSuda:
Still, there are folks interested in Docker for FreeBSD

https://reviews.freebsd.org/D21570
https://github.com/decke/ports/tree/master/sysutils/docker-engine

zhsj:
https://cirrus-ci.org/guide/FreeBSD/ it supports freebsd, and free for public repos.

AkihiroSuda:
containerd/nerdctl uses this package and experimentally supports FreeBSD, so let me close this issue.

The cgo dependency is being dropped in https://github.com/moby/sys/pull/113

PR #32, opened 2020-09-22
Add support for OpenBSD in addition to FreeBSD

thaJeztah:
I am looking at using these modules in containerd, so that we can
reduce the effort of maintaining "mountinfo" implementations.

The containerd variant of this module currently supports OpenBSD in
addition to FreeBSD.

This patch bring in changes similar to the ones made in containerd in:

https://github.com/containerd/containerd/commit/0828b7aa96b55e36a9594e3b98a3f6800bc0734a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 PACKAGES ?= mountinfo mount
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
-	freebsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
+	freebsd/amd64 openbsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
 
 .PHONY: all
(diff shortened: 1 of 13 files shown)


tklauser:
Just noticed that #36 might conflict with this. `x/sys/unix` provides the `MNT_*` flags for OpenBSD as well, so it should be easy to adjust either of the PRs, depending on which one gets merged first.

kolyshkin:
Means we need to close https://github.com/moby/sys/issues/24

kolyshkin:
Wonder if containerd has actual *BSD users, or is the situation similar to moby/moby Solaris support?

AkihiroSuda (approved):

kolyshkin:
Needs a rebase

thaJeztah:
@kolyshkin rebased (and green 👍)

kolyshkin (approved):
LGTM

PR #16, opened 2020-06-23
mountinfo: fix path unescaping

kolyshkin:
Moby PR https://github.com/moby/moby/pull/38977 and containerd PR https://github.com/containerd/containerd/pull/3159
added the handing of escape sequences in mountinfo paths (root and mountpoint fields).

Unfortunately, it also broke the handling of paths containing double
quotes, as it was pointed out in https://github.com/containerd/containerd/issues/4257

The solution is to stop using strconv.Unquote and write our own
specialized function to deal with escape sequences.

Unit tests added.

In addition, unescape the fstype and source fields (the first two after the `-` separator, since kernel escapes those as well.

PR #33, opened 2020-09-22
mountinfo: add test-cases for double quotes in Linux Mountinfo

thaJeztah:
This adds tests for handling paths containing double quotes, and is the equivalent of the tests added in containerd/containerd#4325


thaJeztah:
@kolyshkin @AkihiroSuda PTAL

kolyshkin:
Not sure we need it. Technically, all this is already covered by `TestUnescape` (see #16).

kolyshkin:
As I suggested in https://github.com/containerd/containerd/pull/4325#issuecomment-646331613 I'd rather have unit tests in such case -- and the "unit" here is the unescape function.

PR #34, opened 2020-09-22
mountinfo: use idiomatic naming for fields

thaJeztah:
follow-up to https://github.com/moby/sys/pull/32

I am looking at using these modules in containerd, so that we can reduce the effort of maintaining "mountinfo" implementations.

This change renames the info fields to match the ones used in containerd, which are more "correct" from a Go naming perspective: https://github.com/containerd/containerd/commit/8cd218237b49b41bffa0035393143571ec01b157

This will be a breaking changes for consumers of this package, but we're still pre-v1.0.0, so if we want to make these changes, this is still possible now.

### DIFF
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -46,18 +46,18 @@ type Info struct {
 	// Mountpoint indicates the mount point relative to the process's root.
 	Mountpoint string
 
-	// Opts represents mount-specific options.
-	Opts string
+	// Options represents mount-specific options.
+	Options string
 
 	// Optional represents optional fields.
 	Optional string
 
-	// Fstype indicates the type of filesystem, such as EXT3.
-	Fstype string
+	// FSType indicates the type of filesystem, such as EXT3.
+	FSType string
 
 	// Source indicates filesystem specific information or "none".
 	Source string
 
-	// VfsOpts represents per super block options.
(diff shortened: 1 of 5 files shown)


kolyshkin:
I tend to agree the new names are better, and guess we can convert existing users. :+1:

kolyshkin:
This will only be problematic for repos which use bots to update dependencies (and only if their maintainers ignore CI failures)

kolyshkin:
Can you please rename `FstypeFilter` while at it? I named it this way because the of the field name we're changing.

kolyshkin:
Alternatively, of course, we can modify containerd to expect these fields. I'm not sure which way is better.

thaJeztah:
Rebased 👍 ptal

kolyshkin (approved):
LGTM

We need to cut a release with this change and convert all the users.

kolyshkin:
> Can you please rename FstypeFilter while at it? I named it this way because the of the field name we're changing.

Ah, we forgot about this one, and surely I found it a minute after I cut a release :(

thaJeztah:
> Ah, we forgot about this one, and surely I found it a minute after I cut a release :(

oh! Sorry, good catch, looks like I missed that comment; thanks for fixing in https://github.com/moby/sys/pull/41

PR #35, opened 2020-09-22
mountinfo: minor cleanup

thaJeztah:
just a minor cleanup


kolyshkin (approved):
LGTM

thaJeztah:
Actually, looks like an implementation of parseInfoFile() is missing on freebsd as well,  so I think we could better go for https://github.com/moby/sys/pull/38

PR #38, opened 2020-09-23
Make GetMountsFromReader linux-only

thaJeztah:
closes https://github.com/moby/sys/pull/35

GetMountsFromReader was a wrapper for parseInfoFile on Linux.
All other implementations do not parse `mountinfo`, so had to
stub out the implemetation and omit the actual reader.

(worse, a FreeBSD implementation was missing, causing compile on
that to fail).

This patch makes GetMountsFromReader linux-only, as it was added
to allow benchmarking the `mountinfo` parsing, which only makes
sense on Linux.

### DIFF
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -11,6 +11,14 @@ import (
 	"strings"
 )
 
+// GetMountsFromReader retrieves a list of mounts from the
+// reader provided, with an optional filter applied (use nil
+// for no filter). This can be useful in tests or benchmarks
+// that provide a fake mountinfo data.
+func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
+	return parseInfoFile(reader, f)
+}
+
 func parseInfoFile(r io.Reader, filter FilterFunc) ([]*Info, error) {
 	s := bufio.NewScanner(r)
 	out := []*Info{}
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -1,7 +1,6 @@
 package mountinfo
 
 import (
-	"io"
 	"os"
 )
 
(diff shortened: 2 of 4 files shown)


mountinfo/mountinfo_linux.go
```diff
@@ -11,6 +11,14 @@ import (
 	"strings"
 )
 
+// GetMountsFromReader retrieves a list of mounts from the
+// reader provided, with an optional filter applied (use nil
+// for no filter). This can be useful in tests or benchmarks
+// that provide a fake mountinfo data.
+func GetMountsFromReader(reader io.Reader, f FilterFunc) ([]*Info, error) {
```
thaJeztah, line 18: Now that GetMountsFromReader is effectively just an alias for parseInfoFile, should we just export parseInfoFile on Linux instead?

Looks like it's not used anywhere currently https://grep.app/search?q=mountinfo.GetMountsFromReader

thaJeztah:
@kolyshkin ptal <3

mountinfo/mountinfo_linux.go
kolyshkin: It is used in a few places, so I'd rather not change the name. Surely we can just rename `parseInfoFile` to `GetMountsFromReader`, if this is what you mean

One such place is https://github.com/opencontainers/runc/pull/2599/commits/999ff3f9dbee376bf1db8abe3fedd9594a7ef582, and another is

kolyshkin: ... and another is https://github.com/moby/moby/pull/40694

kolyshkin: Nope, it's not it. I don't remember why I added it but I definitely had a good reason to :-)

kolyshkin: But surely it makes sense to make it Linux-only

kolyshkin: Please see https://github.com/moby/sys/pull/39

PR #39, opened 2020-09-24
mountinfo: make GetMountinfoFromReader Linux-specific

kolyshkin:
... as no other platform provides something like /proc/self/mountinfo.

Closes: https://github.com/moby/sys/pull/38
Closes: https://github.com/moby/sys/pull/35

@thaJeztah PTAL


thaJeztah (approved):
LGTM, thanks!

PR #40, opened 2020-10-01
mount: support FUSE helper

thaJeztah:
When m.Type starts with either fuse. or fuse3, the mount helper binary mount.fuse
or mount.fuse3 is executed.

Similar to the changes made in containerd:

> This is expected to be used by fuse-overlayfs plugin: https://github.com/AkihiroSuda/containerd-fuse-overlayfs
>
> Motivation
>
> The purpose of the containerd fuse-overlayfs snapshotter plugin is to provide
> OverlayFS functionality for rootless mode without depending on the Ubuntu/Debian
> kernel patch.
> Although fuse-overlayfs provides shiftfs functionality and supports CRFS plugin,
> these functionalities are not planned to be supported by the containerd fuse-overlayfs

### DIFF
--- a/mount/unmount_unix.go
+++ b/mount/unmount_unix.go
@@ -2,13 +2,49 @@
 
 package mount
 
-import "golang.org/x/sys/unix"
+import (
+	"os/exec"
+
+	"golang.org/x/sys/unix"
+)
 
 func unmountBare(target string, flags int) error {
 	return unix.Unmount(target, flags)
 }
 
+// fuseSuperMagic is defined in statfs(2)
+const fuseSuperMagic = 0x65735546
+
+func isFUSE(dir string) bool {
+	var st unix.Statfs_t
+	if err := unix.Statfs(dir, &st); err != nil {
+		return false
+	}
+	return st.Type == fuseSuperMagic
+}
+
+// unmountFUSE attempts to unmount using fusermount/fusermount3 helper binary.
+//
+// For FUSE mounts, using these helper binaries is preferred, see:
+// https://github.com/containerd/containerd/pull/3765#discussion_r342083514
+func unmountFUSE(target string) error {
+	var err error
+	for _, helperBinary := range []string{"fusermount3", "fusermount"} {
+		cmd := exec.Command(helperBinary, "-u", target)
+		err = cmd.Run()
+		if err == nil {
+			return nil
+		}
+	}
+	return err
+}
+
 func unmount(target string, flags int) error {
+	if isFUSE(target) {
+		if err := unmountFUSE(target); err == nil {
+			return nil
+		}
+	}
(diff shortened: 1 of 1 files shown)


thaJeztah:
Opening as draft for now, depending on how much sense it will make to replace the mounting part that's in containerd with moby/sys as well. 

Containerd also has some mount/unmount code for Windows, but that may be very container specify, so TBD if we want to keep this repository as "generic" or "container specific"

mount/unmount_unix.go
```diff
@@ -2,13 +2,49 @@
...
+
+	"golang.org/x/sys/unix"
+)
 
 func unmountBare(target string, flags int) error {
 	return unix.Unmount(target, flags)
 }
 
+// fuseSuperMagic is defined in statfs(2)
+const fuseSuperMagic = 0x65735546
```
kolyshkin, line 16: This needs to be added to `golang.org/x/sys/unix`...

mount/unmount_unix.go
```diff
@@ -2,13 +2,49 @@
...
+func unmountFUSE(target string) error {
+	var err error
+	for _, helperBinary := range []string{"fusermount3", "fusermount"} {
+		cmd := exec.Command(helperBinary, "-u", target)
+		err = cmd.Run()
+		if err == nil {
+			return nil
+		}
+	}
+	return err
```
kolyshkin, line 39: Does `fusermount` return an error if `target` is not mounted? If yes, we should filter it out somehow.

mount/unmount_unix.go
```diff
@@ -2,13 +2,49 @@
...
+}
+
 func unmount(target string, flags int) error {
+	if isFUSE(target) {
```
kolyshkin, line 43: Maybe move it to after a normal mount (need to be checked carefully how it would work).

kolyshkin: Otherwise we're making `unmount` do one extra syscall every time.

mount/unmount_unix.go
thaJeztah: Good one, yes

AkihiroSuda:
The mounter implementation seems not committed

thaJeztah:
Ah, yes, I think I started looking at differences between containerd and this repo, and was not finished yet; thought I'd open to at least discuss if this would be in or out of scope for this repo

kolyshkin:
> if this would be in or out of scope for this repo

I think as long as it's not introducing any noticeable overhead to "normal" mounts it's fine to have it. The question is what do we call "noticeable" :)

srstsavage:
Any chance of this moving forward? If I understand correctly this would allow mounting of FUSE filesystems like glusterfs using the docker's `--mount`/`local` driver.

PR #47, opened 2020-10-02
mountinfo: deprecate PidMountInfo

kolyshkin:
Function PidMountInfo is not named correctly (it should be PID not Pid,
and we already have mountinfo in the package name so it's tautological),
and can not accept FilterFunc.

Besides, there are not too many users of this function (I was only able
to found one, see [1]) and users can switch to GetMountsFromReader.

Depreate this function. The plan is to remove it before v1.0.

[1] https://github.com/DataDog/datadog-agent/pull/6495/commits/406d98801434dedc87cab6195079f735d6903d48

### DIFF
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -148,9 +148,14 @@ func parseMountTable(filter FilterFunc) ([]*Info, error) {
 	return GetMountsFromReader(f, filter)
 }
 
-// PidMountInfo collects the mounts for a specific process ID. If the process
-// ID is unknown, it is better to use `GetMounts` which will inspect
-// "/proc/self/mountinfo" instead.
+// PidMountInfo retrieves the list of mounts from a given process' mount
+// namespace. Unless there is a need to get mounts from a mount namespace
+// different from that of a calling process, use GetMounts.
+//
+// This function is Linux-specific.
+//
+// Deprecated: this will be removed before v1; use GetMountsFromReader with
+// opened /proc/<pid>/mountinfo as an argument instead.
 func PidMountInfo(pid int) ([]*Info, error) {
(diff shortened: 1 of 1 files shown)


thaJeztah:
wrt naming; its also worth considering to take the package name itself into account, eg

    mountinfo.ForPID(pid)
    mountinfo.FromPID(pid)
    mountinfo.Get(filter)

(typing on my phone, so just a quick blurb)

kolyshkin:
I didn't want to rename everything :) as we'll have to convert all users and/or deprecate stuff.
This is more about ability to use filters and PID.

But it's a good idea to make names shorter. Let me think about it.

thaJeztah:
Containerd uses `mount.PID()` https://github.com/containerd/containerd/blob/master/mount/mountinfo_linux.go#L137


That said; have we checked if this function is used? ISTR in containerd it was not in use (need to double check), not sure if it's used in runc?

kolyshkin:
One other way to proceed is to deprecate `PidMountinfo`, suggesting to use `GetMountsFromReader` with the first argument being an opened `/proc/self/mountinfo`. Maybe this would be the best way to go.

kolyshkin:
> have we checked if this function is used?

I think I've seen it used once but don't remember where it was :-\

For now I guess what I said [above](https://github.com/moby/sys/pull/47#issuecomment-703799723) is the best way to go

kolyshkin:
> I think I've seen it used once but don't remember where it was :-\

Ah, here's the (only) user I found: https://github.com/DataDog/datadog-agent/pull/6495/commits/406d98801434dedc87cab6195079f735d6903d48

kolyshkin:
@thaJeztah I've change this PR to only deprecate PidMountInfo, no longer introducing a replacement for it. PTAL

thaJeztah (approved):
LGTM

PR #48, opened 2020-10-02
mountinfo: doc fixes

kolyshkin:
1. Lines that start with a space are treated as verbatim text
   (rendered as \<pre\> in HTML). Fix those.

2. There is no sense in having github markdown (like `var`).

This was inspired by the way it looks right now:
* https://pkg.go.dev/github.com/moby/sys/mountinfo@v0.3.1


thaJeztah (approved):
LGTM

kolyshkin:
Checked locally by running godoc server and looking into the output.

thaJeztah:
Ah, sorry, thought I merged this one

PR #50, opened 2020-10-20
mountinfo: correctness improvements

cyphar:
This fixes two correctness issues with mountinfo (one of which caused opencontainers/runc#2647 to be a security regression):

 * Correctly handle IO errors during scanning.
 * Run the filter function after all fields are parsed.

### DIFF
--- a/mountinfo/mountinfo_filters.go
+++ b/mountinfo/mountinfo_filters.go
@@ -6,9 +6,8 @@ import "strings"
 // used to filter out mountinfo entries we're not interested in,
 // and/or stop further processing if we found what we wanted.
 //
-// It takes a pointer to the Info struct (not fully populated,
-// currently only Mountpoint, FSType, Source, and (on Linux)
-// VFSOptions are filled in), and returns two booleans:
+// It takes a pointer to the Info struct (fully populated with all available
+// fields on the GOOS platform), and returns two booleans:
 //
 // skip: true if the entry should be skipped;
 //
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -18,11 +18,9 @@ import (
 func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
 	s := bufio.NewScanner(r)
 	out := []*Info{}
-	var err error
 	for s.Scan() {
-		if err = s.Err(); err != nil {
-			return nil, err
-		}
+		var err error
+
 		/*
 		   See http://man7.org/linux/man-pages/man5/proc.5.html
 
@@ -132,6 +130,9 @@ func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
 		if stop {
 			break
 		}
+	}
+	if err := s.Err(); err != nil {
+		return nil, err
 	}
 	return out, nil
 }
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -72,7 +72,6 @@ func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
 
 		p := &Info{}
 
-		// Fill in the fields that a filter might check
 		p.Mountpoint, err = unescape(fields[4])
 		if err != nil {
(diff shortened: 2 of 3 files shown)


mountinfo/mountinfo_linux.go
```diff
@@ -18,11 +18,9 @@ import (
 func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
```
AkihiroSuda, line 18: Can we have a UT?

mountinfo/mountinfo_linux.go
```diff
@@ -18,11 +18,9 @@ import (
...
 	s := bufio.NewScanner(r)
 	out := []*Info{}
-	var err error
 	for s.Scan() {
-		if err = s.Err(); err != nil {
```
kolyshkin, line 23: That's a curse that I have removed this from so many projects but failed to do so for this one :)

kolyshkin:
> Previously, the filter would be run before all of the fields were parsed
> (this behaviour also was not documented) -- this resulted in users of
> the function accidentally assuming that fields like fsinfo.Root would
> actually be filled correctly.

This is actually documented:

https://github.com/moby/sys/blob/95f2efb64d5a182991b04f79ed60410f25aeabfa/mountinfo/mountinfo_filters.go#L5-L16

so the second commit needs to be amended accordingly.

There is also a comment in the code that needs to be removed:

https://github.com/moby/sys/blob/95f2efb64d5a182991b04f79ed60410f25aeabfa/mountinfo/mountinfo_linux.go#L77

Otherwise LGTM

kolyshkin (changes requested):
Need to update `FilterFunc` docs and remove a comment in the code (see above).

cyphar:
@kolyshkin 

> This is actually documented:

Ah, I was looking at the docs of this function not `FilterFunc`. :bow:

mountinfo/mountinfo_linux.go
thaJeztah: Does golang-ci-lint detect these?

mountinfo/mountinfo_filters.go
```diff
@@ -6,9 +6,8 @@ import "strings"
 // used to filter out mountinfo entries we're not interested in,
 // and/or stop further processing if we found what we wanted.
 //
-// It takes a pointer to the Info struct (not fully populated,
-// currently only Mountpoint, FSType, Source, and (on Linux)
-// VFSOptions are filled in), and returns two booleans:
+// It takes a pointer to the Info struct (fully populated with all available
+// fields on the GOOS platform), and returns two booleans:
```
thaJeztah, line 10: Wondering ; is the "on the GOOS platform" part relevant in the comment? Otherwise we could simplify to;

```suggestion
// It takes a pointer to the populated Info struct, and returns two booleans:
```

mountinfo/mountinfo_linux.go
cyphar: Added. There were a few for this function already, just not one for the `FilterFunc` callback.

mountinfo/mountinfo_filters.go
cyphar: I guess the point I was trying to make is that on FreeBSD, not all of the fields are actually filled because some of the `Info` struct fields are Linux-only concepts.

thaJeztah: Ah, right, and they use the same struct 🤔 gotcha!

It was mostly a suggestion, trying to keep it simple, definitely not a blocker 👍

thaJeztah (approved):
LGTM

mountinfo/mountinfo_linux.go
kolyshkin: I guess not, otherwise that'd be fixed earlier

kolyshkin (approved):
LGTM

thaJeztah:
@kolyshkin should we tag a new release with this?

kolyshkin:
> @kolyshkin should we tag a new release with this?

I already did, https://github.com/moby/sys/releases/tag/mountinfo%2Fv0.4.0

Yet there are no users that are affected (runc was, but this is being fixed by https://github.com/opencontainers/runc/pull/2655)

thaJeztah:
Oh! Somehow didn't get a notification; thanks!

PR #52, opened 2020-10-21
mountinfo: docs nits

kolyshkin:
###    mountinfo: docs nits
    
1. Add a "custom file" use case to GetMountsFromReader.
    
2. Remove obsoleted old comment from parseMountTable.

### mountinfo.Info: refresh doc
    
Refresh documentation about mountinfo.Info fields, based on the latest
proc(5) from man-pages v5.08.

### DIFF
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -29,35 +29,38 @@ type Info struct {
 	// ID is a unique identifier of the mount (may be reused after umount).
 	ID int
 
-	// Parent indicates the ID of the mount parent (or of self for the top of the
-	// mount tree).
+	// Parent is the ID of the parent mount (or of self for the root
+	// of this mount namespace's mount tree).
 	Parent int
 
-	// Major indicates one half of the device ID which identifies the device class.
-	Major int
+	// Major and Minor are the major and the minor components of the Dev
+	// field of unix.Stat_t structure returned by unix.*Stat calls for
+	// files on this filesystem.
+	Major, Minor int
 
-	// Minor indicates one half of the device ID which identifies a specific
-	// instance of device.
-	Minor int
-
-	// Root of the mount within the filesystem.
+	// Root is the pathname of the directory in the filesystem which forms
+	// the root of this mount.
 	Root string
 
-	// Mountpoint indicates the mount point relative to the process's root.
+	// Mountpoint is the pathname of the mount point relative to the
+	// process's root directory.
 	Mountpoint string
 
-	// Options represents mount-specific options.
+	// Options is a comma-separated list of mount options.
 	Options string
 
-	// Optional represents optional fields.
+	// Optional are zero or more fields of the form "tag[:value]",
+	// separated by a space.  Currently, the possible optional fields are
+	// "shared", "master", "propagate_from", and "unbindable". For more
+	// information, see mount_namespaces(7) Linux man page.
 	Optional string
 
-	// FSType indicates the type of filesystem, such as EXT3.
+	// FSType is the filesystem type in the form "type[.subtype]".
 	FSType string
 
-	// Source indicates filesystem specific information or "none".
+	// Source is filesystem-specific information, or "none".
 	Source string
 
-	// VFSOptions represents per super block options.
+	// VFSOptions is a comma-separated list of superblock options.
 	VFSOptions string
 }
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -12,7 +12,8 @@ import (
 // GetMountsFromReader retrieves a list of mounts from the
 // reader provided, with an optional filter applied (use nil
 // for no filter). This can be useful in tests or benchmarks
-// that provide a fake mountinfo data.
+// that provide fake mountinfo data, or when a source other
+// than /proc/self/mountinfo needs to be read from.
 //
 // This function is Linux-specific.
 func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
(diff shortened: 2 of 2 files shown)


mountinfo/mountinfo.go
```diff
@@ -29,35 +29,38 @@ type Info struct {
...
+	// Parent is the ID of the parent mount (or of self for the root
+	// of this mount namespace's mount tree).
 	Parent int
 
-	// Major indicates one half of the device ID which identifies the device class.
-	Major int
+	// Major and Minor are the major and the minor components of the Dev
+	// field of unix.Stat_t structure returned by unix.*Stat calls for
+	// files on this filesystem.
+	Major, Minor int
```
thaJeztah, line 39: Curious; how does the documentation show these combined comments? Does it repeat the same doc for each of them, or does it also should them grouped?

mountinfo/mountinfo.go
```diff
@@ -29,35 +29,38 @@ type Info struct {
...
-	// FSType indicates the type of filesystem, such as EXT3.
+	// FSType is the filesystem type in the form "type[.subtype]".
 	FSType string
 
-	// Source indicates filesystem specific information or "none".
+	// Source is filesystem-specific information, or "none".
 	Source string
 
-	// VFSOptions represents per super block options.
+	// VFSOptions is per-superblock options, comma-separated.
```
thaJeztah, line 64:
```suggestion
	// VFSOptions are per-superblock options, comma-separated.
```

or

```suggestion
	// VFSOptions represents per-superblock options, comma-separated.
```

mountinfo/mountinfo.go
```diff
@@ -29,35 +29,38 @@ type Info struct {
...
+	// the root of this mount.
 	Root string
 
-	// Mountpoint indicates the mount point relative to the process's root.
+	// Mountpoint is the pathname of the mount point relative to the
+	// process's root directory.
 	Mountpoint string
 
-	// Options represents mount-specific options.
+	// Options is per mount options, comma-separated.
```
thaJeztah, line 49:
```suggestion
	// Options are per mount options, comma-separated.
```

or perhaps

```suggestion
	// Options represents per mount options, comma-separated.
```

or `s/represents/stores/` or `s/represents/holds/` ?

mountinfo/mountinfo_linux.go
```diff
@@ -12,7 +12,8 @@ import (
 // GetMountsFromReader retrieves a list of mounts from the
 // reader provided, with an optional filter applied (use nil
 // for no filter). This can be useful in tests or benchmarks
-// that provide a fake mountinfo data.
+// that provide a fake mountinfo data, or when a source other
```
thaJeztah, line 15:
```suggestion
// that provide fake mountinfo data, or when a source other
```

mountinfo/mountinfo.go
kolyshkin: 1. `go doc` shows it as is:
```
	// Major and Minor are the major and the minor components of the Dev
	// field of unix.Stat_t structure returned by unix.*Stat calls for
	// files on this filesystem.
	Major, Minor int
```

2. `go doc` (a webserver) does exactly the same since it's a struct members:

```
type Info
Info reveals information about a particular mounted filesystem. This struct is populated from the content in the /proc/<pid>/mountinfo file.

type Info struct {
    // ID is a unique identifier of the mount (may be reused after umount).
    ID int

    // Parent is the ID of the parent mount (or of self for the root
    // of this mount namespace's mount tree).
    Parent int

    // Major and Minor are the major and the minor components of the Dev
    // field of unix.Stat_t structure returned by unix.*Stat calls for
    // files on this filesystem.
    Major, Minor int
...
```

mountinfo/mountinfo.go
kolyshkin: I don't like the term "represents" here (as it implies some kind of conversion).

I don't like "stores" or "holds", as this is a string, and yes, it stores some value but it's kinda obvious.

Brevity is good here.

mountinfo/mountinfo.go
kolyshkin: Not a native speaker, but `VFSOptions` is a (single) variable, so AFAICS using plural is wrong here, as we don't describe the "options", but a variable.

kolyshkin: Also, if it would be plural, we'd say "VFSOptions represent". See, this feels wrong -- as we don't talk about options (plural), but a variable (singular).

mountinfo/mountinfo_linux.go
kolyshkin: yep, fixed!

kolyshkin:
@thaJeztah addressed your comments, PTAL

mountinfo/mountinfo.go
thaJeztah: Ah, yes, had wires crossed and thought we were in a `var()` block 🤦 

makes sense

mountinfo/mountinfo.go
thaJeztah: > we don't describe the "options", but a variable.
> Also, if it would be plural, we'd say "VFSOptions represent". See, this feels wrong -- as we don't talk about options (plural), but a variable (singular).

Yes, you're right; it should be describing the variable, not the options in the variable.

Something feels "off" in using `is` here, and I think I just found the problem/solution; usually (in case of a "slice"); how about this:

```suggestion
	// VFSOptions is a comma-separated list of per-superblock options.
```

Perhaps the `per-` could be dropped (?)

```suggestion
	// VFSOptions is a comma-separated list of superblock options.
```

mountinfo/mountinfo.go
thaJeztah: see my other comment, I think we can use the same approach here;

```suggestion
	// Options is a comma-separated list of per mount options.
```

Could perhaps omit the `per` here (?);

```suggestion
	// Options is a comma-separated list of mount options.
```

thaJeztah (changes requested):

kolyshkin:
The wording ("per-superblock" and "per-mount") was taken from proc(5) man page, but I have changed it according to last review from @thaJeztah. PTAL

thaJeztah:
> The wording ("per-superblock" and "per-mount") was taken from proc(5) man page

Ah! Didn't know. I guess removing it is ok (if we hear it's not clear we could always add back the `per-`)

thaJeztah (approved):
LGTM, thanks!

PR #53, opened 2020-10-29
Merge pkg/truncindex and pkg/symlink from moby/moby

dims:
We need to remove the recursive dependency of containerd<->moby/moby, containerd uses 2 packages from moby/moby. So to break the recursion, we need to move those two packages into a "neutral" repo. per the discussion in https://github.com/containerd/containerd/pull/4631, the idea was to preserve the git history of these two packages and merge them into this repository

cc @thaJeztah

### DIFF
--- a/pkg/symlink/LICENSE.APACHE
+++ b/pkg/symlink/LICENSE.APACHE
@@ -1,0 +1,191 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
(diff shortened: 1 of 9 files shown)


dims:
Notes on how the git history was pruned and grafted:
```
~/go/src/github.com/moby/moby> git checkout -b merging-some-branches
~/go/src/github.com/moby/moby> git filter-repo --path pkg/symlink --path pkg/truncindex
~/go/src/github.com/moby/moby> cd ../sys
~/go/src/github.com/moby/moby> git remote add patchy ../moby
~/go/src/github.com/moby/moby> git fetch patchy
~/go/src/github.com/moby/moby> git merge --allow-unrelated-histories --signoff -m "Merge pkg/(symlink|truncindex) from moby/moby into moby/sys" patchy/merging-some-branches
```

dims:
@thaJeztah i don't feel that i should not try to "fix" the DCO failure, so leaving it as-is.

thaJeztah:
Oh, thanks! Erm (oh, feel bad commenting that, because I didn't come round to try, and you did all the work), I think the idea was to only merge `pkg/symlink` here, and for `pkg/truncindex`, just creating a copy in containerd would be ok (could of course keep history there as well if useful)

Thanks for providing the steps you took though; let me give it a quick try to repeat with pkg/symlink only

thaJeztah:
> i don't feel that i should not try to "fix" the DCO failure, so leaving it as-is.

Agreed; we can't change history, so no need to fix-up.

Looks like most of them are because of the old `Docker-DCO..` that we used in the past, although the GitHub handle in the first one I picked looks odd (possibly because it was a cherry-pick / carry);

```
Docker-DCO-1.1-Signed-off-by: Michael Crosby <michael@docker.com> (github: vieux)
```

There's also some where it complains about e-mail addresses, wonder if it supports `.mailmap` (not worth the effort though)

dims:
thanks @thaJeztah ! the new PR looks good! ( i had missed a couple of things like stripping out `pkg/` and adding the new package to the Makefile etc. )

thaJeztah:
Thanks for starting the work; your steps helped me getting it done (without having to do a lot of Google-fu 😂)

PR #54, opened 2020-10-29
Migrate github.com/docker/docker/pkg/symlink

thaJeztah:
replaces https://github.com/moby/sys/pull/53
closes https://github.com/moby/sys/pull/53
relates to https://github.com/containerd/containerd/pull/4631

Strategy taken:

```bash
# install filter-repo (https://github.com/newren/git-filter-repo/blob/main/INSTALL.md)
brew install git-filter-repo

cd ~/projects

# create a temporary clone of docker
git clone https://github.com/docker/docker.git moby_symlink
cd moby_symlink

# create branch to work with
git checkout -b migrate_pkg_symlink

# remove all code, except for pkg/symlink
git filter-repo --path pkg/symlink

# go to the target github.com/moby/sys repository
cd ~/projects/moby-sys

# create a branch to work with
git checkout -b integrate_moby_symlink

# add the temporary repository as an upstream and make sure it's up-to-date
git remote add moby_symlink ~/projects/moby_symlink
git fetch moby_symlink

# merge the upstream code, rewriting "pkg/symlink" to "symlink"
git merge --allow-unrelated-histories --signoff -S -Xsubtree=pkg/ moby_symlink/migrate_pkg_symlink
```

After the above;

- symlink: remove "github.com/docker/docker" import enforcement comments
- symlink: add go.mod
- Makefile: run tests on symlink module
- symlink: add package doc

PR #55, opened 2020-11-05
symlink: drop go 1.13, and use testing.Cleanup()

thaJeztah:
Cleanup was added in Go 1.14, so we need to drop go 1.13 for this.

### DIFF
--- a/symlink/fs_unix_test.go
+++ b/symlink/fs_unix_test.go
@@ -53,8 +53,8 @@ func testSymlink(tmpdir, path, expected, scope string) error {
 }
 
 func TestFollowSymlinkAbsolute(t *testing.T) {
-	tmpdir, cleanup := mkTempDir(t)
-	defer cleanup()
+	tmpdir := mkTempDir(t)
+
 	if err := makeFs(tmpdir, []dirOrLink{{path: "testdata/fs/a/d", target: "/b"}}); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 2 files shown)


thaJeztah:
relates to https://github.com/moby/sys/pull/54#discussion_r517726553

/cc @kolyshkin

symlink/fs_unix_test.go
```diff
@@ -53,8 +53,8 @@ func testSymlink(tmpdir, path, expected, scope string) error {
 }
 
 func TestFollowSymlinkAbsolute(t *testing.T) {
-	tmpdir, cleanup := mkTempDir(t)
-	defer cleanup()
+	tmpdir := mkTempDir(t)
```
kolyshkin, line 56: Why don't we use `t.TempDir()` here? https://golang.org/pkg/testing/#T.TempDir

Ahh, this is because they are only supported since 1.15, so we need to wait at least until 1.14 is no longer supported.

Makes sense to add a TODO maybe?

thaJeztah: Ah, yes, can add a TODO

(didn't know about the `t.TempDir()` until you mentioned it recently. I  should catch up more on new features 😂)

kolyshkin (changes requested):
LGTM except for a missing TODO

symlink/fs_unix_test.go
kolyshkin: Can't use this shiny new t.TempDir() anyway until 1.14 goes out of fashion...

kolyshkin (approved):
LGTM

kolyshkin:
@cpuguy83 PTAL

kolyshkin:
@cpuguy83 PTAL

cpuguy83 (approved):
LGTM

PR #59, opened 2020-12-18
Create proper mounter for OpenBSD

yukiisbored:
This is made for the sysutils/docker-cli port which I maintain, as moby/sys is part of the vendored libraries for docker/cli.

While I could stub this out (the same way as how it is on darwin and windows), I decided to actually write a proper mounter since it seems there's intention to actually bring OpenBSD support (as seen in https://github.com/moby/sys/commit/cb807b1e4cf01065587739df73d758d6d171819b).

OpenBSD's [mount(2)](https://man.openbsd.org/mount.2) works differently since it uses separate structs to present options for different filesystems, unlike in FreeBSD. It also means that support for other filesystems need to be added explicitly. I'm not sure which ones are required for now so I just write support for only FFS, which is the go-to filesystem used by OpenBSD systems.

### DIFF
--- a/mount/mounter_openbsd.go
+++ b/mount/mounter_openbsd.go
@@ -1,0 +1,77 @@
+// +build openbsd,cgo
+
+/*
+   Due to how OpenBSD mount(2) works, filesystem types need to be
+   supported explicitly since it uses separate structs to pass
+   filesystem-specific arguments.
+
+   For now only UFS/FFS is supported as it's the default fs
+   on OpenBSD systems.
+
+   See: https://man.openbsd.org/mount.2
+*/
+
+package mount
+
+/*
+#include <sys/types.h>
+#include <sys/mount.h>
+*/
+import "C"
+
+import (
+	"fmt"
+	"syscall"
+	"unsafe"
+)
+
+func createExportInfo(readOnly bool) C.struct_export_args {
+	exportFlags := C.int(0)
+	if readOnly {
+		exportFlags = C.MNT_EXRDONLY
+	}
+	out := C.struct_export_args{
+		ex_root:  0,
+		ex_flags: exportFlags,
+	}
+	return out
+}
+
+func createUfsArgs(device string, readOnly bool) unsafe.Pointer {
+	out := &C.struct_ufs_args{
+		fspec:       C.CString(device),
+		export_info: createExportInfo(readOnly),
+	}
+	return unsafe.Pointer(out)
+}
+
+func mount(device, target, mType string, flag uintptr, data string) error {
+	readOnly := flag&RDONLY != 0
+
+	var fsArgs unsafe.Pointer
+
+	switch mType {
+	case "ffs":
+		fsArgs = createUfsArgs(device, readOnly)
+	default:
+		return &mountError{
+			op:     "mount",
+			source: device,
+			target: target,
+			flags:  flag,
+			err:    fmt.Errorf("unsupported file system type: %s", mType),
+		}
+	}
+
+	if errno := C.mount(C.CString(mType), C.CString(target), C.int(flag), fsArgs); errno != 0 {
+		return &mountError{
+			op:     "mount",
+			source: device,
+			target: target,
+			flags:  flag,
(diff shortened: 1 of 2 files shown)


thaJeztah (changes requested):
Thanks!

Could you perhaps include this as the commit message ? (Perhaps also a note/comment in the code to explain that more filesystems can be added in future, but I'm ok with skipping that if it's adding too much noise in the code)

```
OpenBSD's mount(2) works differently since it uses separate structs to present
options for different filesystems, unlike in FreeBSD. It also means that support
for other filesystems need to be added explicitly. I'm not sure which ones are
required for now so I just write support for only FFS, which is the go-to
filesystem used by OpenBSD systems.
```

thaJeztah (approved):
LGTM, thanks!

thaJeztah:
@cpuguy83 @kolyshkin PTAL

mount/mounter_openbsd.go
```diff
@@ -0,0 +1,76 @@
...
+func createUfsArgs(device string, readOnly bool) unsafe.Pointer {
+	out := &C.struct_ufs_args{
+		fspec:       C.CString(device),
+		export_info: createExportInfo(readOnly),
+	}
+	return unsafe.Pointer(out)
+}
+
+func mount(device, target, mType string, flag uintptr, data string) error {
+	readOnly := flag&RDONLY == 1
```
kolyshkin, line 49: Strictly speaking, this should be
```suggestion
	readOnly := flag&RDONLY == RDONLY
```
or
```suggestion
	readOnly := flag&RDONLY != 0
```

mount/mounter_openbsd.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if mType == "ffs" {
+		fsArgs = createUfsArgs(device, readOnly)
+	} else {
+		return &mountError{
+			op:     "mount",
+			source: device,
+			target: target,
+			flags:  flag,
+			err:    fmt.Errorf("unsupported type"),
```
kolyshkin, line 61: `s/fmt.Errorf/errors.New/` as you don't use any %-style formatting

mount/mounter_openbsd.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	return unsafe.Pointer(out)
+}
+
+func mount(device, target, mType string, flag uintptr, data string) error {
+	readOnly := flag&RDONLY == 1
+
+	var fsArgs unsafe.Pointer
+
+	if mType == "ffs" {
```
kolyshkin, line 53: Since we might support more filesystems in the future, I'd use `switch/case` here.

mount/mounter_openbsd.go
kolyshkin: or
```go
fmt.Errorf("unsupported filesystem type: %s", mType)
```

kolyshkin:
@yukiisbored thanks for your contribution, I left a few review comments.

thaJeztah:
@kolyshkin looks like the PR was updated; PTAL

mount/mounter_openbsd.go
```diff
@@ -0,0 +1,77 @@
...
+	switch mType {
+	case "ffs":
+		fsArgs = createUfsArgs(device, readOnly)
+	default:
+		return &mountError{
+			op:     "mount",
+			source: device,
+			target: target,
+			flags:  flag,
+			err:    fmt.Errorf("unsupported file system type: %s", mType),
```
kolyshkin, line 62: I'd use string concatenation here (it's a bit faster):
```suggestion
			err:    errors.New("unsupported file system type: " + mType),
```

but feel free to ignore (not that this is a hot path)

kolyshkin (approved):
LGTM

mount/mounter_openbsd.go
thaJeztah: 😉😇 https://github.com/moby/sys/pull/59#discussion_r551566848

thaJeztah:
Thanks!!

PR #60, opened 2020-12-28
Add support for NetBSD

iamleot:
This add support for NetBSD (noticed when building docker/cli 20.10.1).


iamleot:
That's probably not okay: it fixes the build issue for docker/cli but still not implement any features.

Closing until it's properly implemented.

PR #61, opened 2021-01-18
mountinfo: fix PrefixFilter() being too greedy

thaJeztah:
The PrefixFilter matched on a literal "prefix", making it too greedy, and,
for example, filtering on "/a" prefix would also return mounts for
"/aaa" or "/abc".

This patch changes the matching to match on a prefix / parent _path_ instead.

### DIFF
--- a/mountinfo/mountinfo_filters.go
+++ b/mountinfo/mountinfo_filters.go
@@ -14,11 +14,16 @@ import "strings"
 // stop: true if parsing should be stopped after the entry.
 type FilterFunc func(*Info) (skip, stop bool)
 
-// PrefixFilter discards all entries whose mount points
-// do not start with a specific prefix.
+// PrefixFilter discards all entries whose mount points do not start with, or
+// are equal to the path specified in prefix. The prefix path must be absolute,
+// have all symlinks resolved, and cleaned (i.e. no extra slashes or dots).
+//
+// PrefixFilter treats prefix as a path, not a partial prefix, which means that
+// given "/foo", "/foo/bar" and "/foobar" entries, PrefixFilter("/foo") returns
+// "/foo" and "/foo/bar", and discards "/foobar".
 func PrefixFilter(prefix string) FilterFunc {
 	return func(m *Info) (bool, bool) {
-		skip := !strings.HasPrefix(m.Mountpoint, prefix)
+		skip := !strings.HasPrefix(m.Mountpoint+"/", prefix+"/")
 		return skip, false
 	}
 }
--- a/mountinfo/mountinfo_filters_test.go
+++ b/mountinfo/mountinfo_filters_test.go
@@ -1,0 +1,31 @@
+package mountinfo
+
+import "testing"
+
+func TestPrefixFilter(t *testing.T) {
+	tests := []struct {
+		prefix     string
+		mountPoint string
+		shouldSkip bool
+	}{
+		{prefix: "/a", mountPoint: "/a", shouldSkip: false},
+		{prefix: "/a", mountPoint: "/a/b", shouldSkip: false},
+		{prefix: "/a", mountPoint: "/aa", shouldSkip: true},
+		{prefix: "/a", mountPoint: "/aa/b", shouldSkip: true},
+
+		// invalid prefix: prefix path must be cleaned and have no trailing slash
+		{prefix: "/a/", mountPoint: "/a", shouldSkip: true},
+		{prefix: "/a/", mountPoint: "/a/b", shouldSkip: true},
+	}
+	for _, tc := range tests {
+		filter := PrefixFilter(tc.prefix)
+		skip, _ := filter(&Info{Mountpoint: tc.mountPoint})
+		if skip != tc.shouldSkip {
+			if tc.shouldSkip {
+				t.Errorf("prefix %q: expected %q to be skipped", tc.prefix, tc.mountPoint)
+			} else {
+				t.Errorf("prefix %q: expected %q not to be skipped", tc.prefix, tc.mountPoint)
+			}
+		}
+	}
+}


mountinfo/mountinfo_filters.go
```diff
@@ -18,7 +18,7 @@ type FilterFunc func(*Info) (skip, stop bool)
 // do not start with a specific prefix.
 func PrefixFilter(prefix string) FilterFunc {
 	return func(m *Info) (bool, bool) {
-		skip := !strings.HasPrefix(m.Mountpoint, prefix)
+		skip := m.Mountpoint != prefix && !strings.HasPrefix(m.Mountpoint, prefix+"/")
```
thaJeztah, line 21: I'm a bit in doubt if matching on a literal prefix was a valid use-case; if so, perhaps we need a new filter (e.g., `ParentFilter` or `ParentPathFilter`) ?

thaJeztah:
@kolyshkin @cpuguy83 ptal

If this is indeed a bug, we should also fix `ParentsFilter`, which uses similar logic

mountinfo/mountinfo_filters.go
thaJeztah: We should also consider wether or not `prefix` itself should be included or not (i.e., if it's _inclusive_ of `prefix`).

If not, I need to remove `m.Mountpoint != prefix` from the above;

```suggestion
		skip := !strings.HasPrefix(m.Mountpoint, prefix+"/")
```

thaJeztah:
@kolyshkin @cpuguy83 ptal 🤗

mountinfo/mountinfo_filters.go
kolyshkin: For the second question (whether `PrefixFilter` should return the mountpoint if it equals to prefix) -- I believe it should (example, if I use `PrefixFilter("/usr")` and the `/usr` itself is a mount point, I'd expect it to be included.

kolyshkin:
I would leave this as-is, instead improving the documentation, saying something like

```
Note that the argument is used as-is.

Example 1: PrefixFilter("/a") will match directories like "/aa" and "/aaa".

Example 2: PrefixFilter("/usr") will match "/usr" and "/usr/local", while
PrefixFilter("/usr/") won't match "/usr".
```

or something like this.

IOW, I think there's no need to modify the code, but maybe it makes sense
to explain how it works. In case something like what you did in this PR is needed,
user can add a `/` suffix to the argument.

kolyshkin:
> If this is indeed a bug, we should also fix ParentsFilter, which uses similar logic

At the moment I can't think of a scenario in which `ParentsFilter` can give false
positives (i.e. return a directory which is not a parent of the argument).

If you can, could you please share?

thaJeztah:
> At the moment I can't think of a scenario in which ParentsFilter can give false positives (i.e. return a directory which is not a parent of the argument).

No I don't have examples where this has lead to issues. That said, it's very easy to accidentally strip trailing directory separators, e.g. https://play.golang.org/p/uttCSSgRtsJ

```go
package main

import (
	"fmt"
	"path"
)

func main() {
	foo1 := "/dev/foo/"
	foo2 := path.Clean(foo1)
	prefix1 := path.Dir(foo1)
	prefix2 := path.Dir(foo2)

	fmt.Println(foo1)
	fmt.Println(foo2)
	fmt.Println(prefix1)
	fmt.Println(prefix2)
}
```

Getting back to;

> In case something like what you did in this PR is needed, user can add a / suffix to the argument.

(thinking out lout); _are_ there use-cases (in current use of this function) where you'd want `PrefixFilter(/foo)` to return `/foo`, `/foo1`, `/foo2` ?

So, _technically_ the functionality is in the name (it's filtering on "prefix", not on "path"), but all uses I currently see, the intent _looks to be_  `PrefixFilter(/foo/) + SingleEntryFilter(/foo)`.

As an example, `mount.RecursiveUnmount(/foo)` will currently unmount `/foo`, `/foo/*`, `/foobar`, `/foobar/*` etc., which may be unexpected.

mountinfo/mountinfo_filters.go
kolyshkin: OK, I think the check here should be
```suggestion
skip := !strings.HasPrefix(m.Mountpoint+"/", prefix+"/")
```

kolyshkin:
> As an example, mount.RecursiveUnmount(/foo) will currently unmount /foo, /foo/*, /foobar, /foobar/* etc., which may be unexpected.

OK, this is definitely a bug. You're right about how easy it is to loose the ending slash, too.

SGTM with the change suggested in https://github.com/moby/sys/pull/61#discussion_r577260268, and an update to the filter doc.

kolyshkin:
Suggested doc change:

```go
// PrefixFilter discards all entries whose mount points
// do not start with a specific prefix. The prefix is treated
// as an absolute directory path (i.e. "/mnt/dir" argument
// will match "/", "/mnt", and "/mnt/dir" entries, but won't
// match e.g. "/mnt/dir2").
```

thaJeztah:
>> As an example, mount.RecursiveUnmount(/foo) will currently unmount /foo, /foo/, /foobar, /foobar/ etc., which may be unexpected.
> OK, this is definitely a bug. You're right about how easy it is to loose the ending slash, too.

Keeping myself honest, I should add that that won't happen if the fast path succeeds successfully. which does an `unmount` on the path itself https://github.com/moby/sys/blob/13322f14efa69fa3b46946bc2374e811613c1bc4/mount/mount_unix.go#L43-L51

> SGTM with the change suggested in #61 (comment), and an update to the filter doc.

Thanks! Let me have a look at that suggestion; I recall looking at my changes yesterday and think I made a mistake for situations where the given path already has a `/` (in which case it would check for prefix `/path//` with double trailing slashes)

kolyshkin:
> Thanks! Let me have a look at that suggestion; I recall looking at my changes yesterday and think I made a mistake for situations where the given path already has a / (in which case it would check for prefix /path// with double trailing slashes)

The doc already says the path should be clean (even the extra slash is explicitly mentioned).

https://github.com/moby/sys/blob/5a29239ab7e72fef7786362a4f0ff2fd4459e759/mountinfo/doc.go#L18-L19

kolyshkin:
> The doc already says the path should be clean

...but we can re-emphasize it again in PrefixFilter doc.

mountinfo/mountinfo_filters_test.go
```diff
@@ -0,0 +1,36 @@
...
+		{prefix: "/a", mountPoint: "/a/b/", shouldSkip: false},
+		{prefix: "/a", mountPoint: "/aa", shouldSkip: true},
+		{prefix: "/a", mountPoint: "/aa/", shouldSkip: true},
+		{prefix: "/a", mountPoint: "/aa/b", shouldSkip: true},
+		{prefix: "/a", mountPoint: "/aa/b/", shouldSkip: true},
+
+		// invalid prefix: prefix path must be cleaned and have no trailing slash
+		{prefix: "/a/", mountPoint: "/a", shouldSkip: true},
+		{prefix: "/a/", mountPoint: "/a/b", shouldSkip: true},
+		{prefix: "/a/", mountPoint: "/a/b/", shouldSkip: true},
```
thaJeztah, line 23: @kolyshkin I made the change you suggested in https://github.com/moby/sys/pull/61#discussion_r577260268, but I kept the `m.Mountpoint != prefix`. I wasn't sure if mountinfo could return paths with a trailing slash (at least we don't do a `path.Clean()` in our code.

If we're sure mountinfo will _never_ return a trailing slash, then we can remove some of these tests;

```suggestion
		{prefix: "/a", mountPoint: "/a", shouldSkip: false},
		{prefix: "/a", mountPoint: "/a/b", shouldSkip: false},
		{prefix: "/a", mountPoint: "/aa", shouldSkip: true},
		{prefix: "/a", mountPoint: "/aa/b", shouldSkip: true},

		// invalid prefix: prefix path must be cleaned and have no trailing slash
		{prefix: "/a/", mountPoint: "/a", shouldSkip: true},
		{prefix: "/a/", mountPoint: "/a/b", shouldSkip: true},
```

mountinfo/mountinfo_filters.go
```diff
@@ -14,11 +14,16 @@ import "strings"
...
 
-// PrefixFilter discards all entries whose mount points
-// do not start with a specific prefix.
+// PrefixFilter discards all entries whose mount points do not start with, or
+// are equal to the path specified in prefix. The prefix path must be absolute,
+// have all symlinks resolved, and cleaned (i.e. no extra slashes or dots).
+//
+// PrefixFilter treats prefix as a path, not a partial prefix, which means that
+// given "/foo", "/foo/bar" and "/foobar" entries, PrefixFilter("/foo") returns
+// "/foo" and "/foo/bar", and discards "/foobar".
```
thaJeztah, line 23: Technically, this function is now Linux/Unix-specific (well, of course all of these filters have no meaning on Windows); should we move rename this file (except for the `type FilterFunc`) to `mountinfo_filters_unix.go` and exclude windows?

mountinfo/mountinfo_filters_test.go
kolyshkin: I am pretty sure mountinfo never returns something with an ending slash; IOW we can assume that the paths it provides are as clean as `filepath.Clean()` makes them.

mountinfo/mountinfo_filters.go
kolyshkin: Makes sense

kolyshkin:
> I kept the `m.Mountpoint != prefix`.

I might need more :coffee: at the moment, but I think there's no point in keeping this check.

mountinfo/mountinfo_filters_test.go
thaJeztah: Thx; I'll remove those case

thaJeztah:
> I might need more ☕ at the moment, but I think there's no point in keeping this check.

I'll probably need more ☕ as well,  don't recall now which cases it was needed for, but could've been for the "does mountinfo ever return a trailing slash".

Of course, it still saves a call to `strings.HasPrefix()` (doing just a straight compare first as fast path)

kolyshkin:
> Of course, it still saves a call to strings.HasPrefix() (doing just a straight compare first as fast path)

This is basically the same as `HasPrefix`, which is implemented as:
```go
func HasPrefix(s, prefix string) bool {
    	return len(s) >= len(prefix) && s[0:len(prefix)] == prefix
```
and in case `m.Mountpoint == prefix` your code will result in two string comparisons rather than one (and a real string comparison (in Go that means if len is the same) is relatively costly).

I might be missing something, it looks like this optimization does not make sense. I tried to come up with a benchmark but it takes too much time...

kolyshkin:
Oh, and add commits from https://github.com/moby/sys/pull/64 to this PR if it makes sense to you.

kolyshkin:
Ah, out of stupid curiosity I wrote a benchmark:

```go
package main

import (
	"strings"
	"testing"

	"github.com/moby/sys/mountinfo"
)

var (
	mounts   []string
	prefixes = []string{
		"whatever",
		"/some/very/long/nonexistent/directory",
		"/another/nonexistent/dir",
	}
)

func init() {
	m, err := mountinfo.GetMounts(nil)
	if err != nil {
		panic(err)
	}
	for _, m := range m {
		mounts = append(mounts, m.Mountpoint)
	}
	prefixes = append(prefixes, mounts...)
}

func benchmarkSkip(b *testing.B, skipFn func(m, p string) bool) {
	var skip bool
	for n := 0; n < b.N; n++ {
		for _, m := range mounts {
			for _, p := range prefixes {
				skip = m != p && !strings.HasPrefix(m+"/", p+"/")
			}
		}
	}
	b.Log(skip)
}

func skip1(m, p string) bool {
	return m != p && !strings.HasPrefix(m+"/", p+"/")
}

func skip2(m, p string) bool {
	return !strings.HasPrefix(m+"/", p+"/")
}

func TestSkipsAreSame(t *testing.T) {
	for _, m := range mounts {
		for _, p := range prefixes {
			if skip1(m, p) != skip2(m, p) {
				t.Errorf("different results for %s, %s", m, p)
			}
		}
	}
}

func BenchmarkSkip1(b *testing.B) {
	benchmarkSkip(b, skip1)
}

func BenchmarkSkip2(b *testing.B) {
	benchmarkSkip(b, skip2)
}
```

and it shows that
1. skip1 and skip2 produce the same results;
2. skip1 and skip2 have the same performance (within the margin of a measurement error -- in most runs skip2 is a _tad_ faster).

Here's a run:
```console
[kir@kir-rhat skip-bench]$ go test -v -bench . -benchtime=4s
=== RUN   TestSkipsAreSame
--- PASS: TestSkipsAreSame (0.00s)
goos: linux
goarch: amd64
pkg: github.com/kolyshkin/test/skip-bench
BenchmarkSkip1
    a_test.go:39: false
    a_test.go:39: false
    a_test.go:39: false
    a_test.go:39: false
BenchmarkSkip1-4   	   56517	     84676 ns/op
BenchmarkSkip2
    a_test.go:39: false
    a_test.go:39: false
    a_test.go:39: false
    a_test.go:39: false
BenchmarkSkip2-4   	   57717	     83108 ns/op
PASS
ok  	github.com/kolyshkin/test/skip-bench	11.284s
```

kolyshkin:
@thaJeztah PTAL

thaJeztah:
oh! made the changes locally, and forgot to push 😂 

updated 👍

kolyshkin (approved):
LGTM

kolyshkin:
@cpuguy83 @AkihiroSuda PTAL

kolyshkin:
Once this is in we need to cut a (minor?) release.

thaJeztah:
yes, guess that won't hurt

cpuguy83 (approved):
LGTM

PR #62, opened 2021-01-18
mount: implement UnmountAll()

thaJeztah:
`UnmountAll()` unmounts all mounts underneath the specified path, but (contrary to `RecursiveUnmount()`) does not unmount `path` itself.

relates to https://github.com/moby/moby/issues/39267 https://github.com/moby/moby/issues/39267#issuecomment-762282680

### DIFF
--- a/mount/mount_unix.go
+++ b/mount/mount_unix.go
@@ -4,7 +4,9 @@ package mount
 
 import (
 	"fmt"
+	"path"
 	"sort"
+	"strings"
 
 	"github.com/moby/sys/mountinfo"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


kolyshkin:
Perhaps `UnmountAllUnder` or `UnmountAllBelow`, to better say what it does.

mount/mount_unix.go
```diff
@@ -85,3 +108,20 @@ func RecursiveUnmount(target string) error {
...
+	// On Linux, mntDetach flag ensures a recursive unmount.  For other
+	// platforms, if there are submounts, we'll get EBUSY (and fall back
+	// to the slow path). NOTE we do not ignore EINVAL here as target might
+	// not be a mount point itself (but there can be mounts underneath).
+	if err := unix.Unmount(target, mntDetach); err == nil {
+		return nil
+	}
+
+	// Slow path: unmount all mounts inside target one by one.
+	return UnmountAll(target)
```
kolyshkin, line 126: It seems that you need to unmount the target itself here.

kolyshkin: If I am not mistaken above, this function now works incorrectly on its slow path, and thus our tests are not adequate.

Surely there needs to be a test added for `UnmountAll`.

mount/mount_unix.go
```diff
@@ -37,35 +39,56 @@ func Unmount(target string) error {
...
+	// Fast path: try to unmount top-level mounts first. This works if target is
+	// a mount point that can be unmounted.
 	// On Linux, mntDetach flag ensures a recursive unmount.  For other
 	// platforms, if there are submounts, we'll get EBUSY (and fall back
-	// to the slow path). NOTE we do not ignore EINVAL here as target might
-	// not be a mount point itself (but there can be mounts underneath).
-	if err := unix.Unmount(target, mntDetach); err == nil {
-		return nil
+	// to the slow path). We're not using RecursiveUnmount() here, to avoid
+	// repeatedly calling mountinfo.GetMounts()
```
kolyshkin, line 55: I think this comment is no longer valid (as RecursiveUnmount itself is using this function).

mount/mount_unix.go
```diff
@@ -37,35 +39,56 @@ func Unmount(target string) error {
...
+			if !strings.HasPrefix(m.Mountpoint, p) {
+				subMounts = append(subMounts, m)
+			}
+		}
 	}
 
 	// Make the deepest mount be first
-	sort.Slice(mounts, func(i, j int) bool {
-		return len(mounts[i].Mountpoint) > len(mounts[j].Mountpoint)
+	sort.Slice(subMounts, func(i, j int) bool {
```
kolyshkin, line 83: I'd add something like
```go
if len(subMounts) == 0 {
     return nil
}
```
here

mount/mount_unix.go
```diff
@@ -37,35 +39,56 @@ func Unmount(target string) error {
 	}
 }
 
-// RecursiveUnmount unmounts the target and all mounts underneath, starting
-// with the deepest mount first. The argument does not have to be a mount
-// point itself.
-func RecursiveUnmount(target string) error {
-	// Fast path, works if target is a mount point that can be unmounted.
+// UnmountAll unmounts all mounts and submounts underneath parent,
```
kolyshkin, line 42:
```suggestion
// UnmountAll unmounts all mounts underneath parent, excluding the parent itself.
// Note that parent does not have to be a mount point.
```

kolyshkin: (I've removed "and submount" because from user's perspective all mounts are the same -- it is only in the code we distinguish between the top-level mounts and their sub-mounts).

mount/mount_unix.go
```diff
@@ -37,35 +39,56 @@ func Unmount(target string) error {
...
-	// not be a mount point itself (but there can be mounts underneath).
-	if err := unix.Unmount(target, mntDetach); err == nil {
-		return nil
+	// to the slow path). We're not using RecursiveUnmount() here, to avoid
+	// repeatedly calling mountinfo.GetMounts()
+
+	var skipParents []string
+	for _, m := range mounts {
+		// Skip parent itself, and skip non-top-level mounts
+		if m.Mountpoint == parent || path.Dir(m.Mountpoint) != parent {
```
kolyshkin, line 60: So, I think this is not entirely correct optimization. Let me describe what I think would be a correct one.

We have a list of mounts, it's a tree, and we assume that if we use `mntDetach` flag we will only need to unmount the top ones. But the top ones are not those whose parent is `parent`, but those who have submounts.

For example, if we have the following directory tree (mounts are marked with `[n]`):

```
.
├── [1]
│   ├── [4]
│   └── [5]
├── 2
│   └── [6]
└── 3
    └── 7
        └── [8]
```

the top-level mounts are [1], [6], and [8], and we should try unmounting them first, hoping that [4] and [5] will be unmounted.

One way of doing that is to sort the tree so the shortest paths will be first, and upon successful unmount remove the children of the unmounted directory.

I am not that great at algorithms and not sure if it's possible to create the one described above, which will be faster that just iterating over a list and unmounting everything. Maybe it will be possible and even efficient if we use a real tree-like structure and do a width-first tree traversal, discarding the entire branch underneath if umount is successful.

My gut feeling though is that the "smart" approach will be slower than the "stupid" one.

kolyshkin: One other things is, for special cases, when we know that the top mounts are right below the parent (which is what your code does in fast path), we can avoid reading mountinfo and just do something like
```go
entries, err := os.Readdirnames(parent)
if err != nil {
   return err
}
for _, de := range entries {
   mp := filepath.Join(parent, de)
   err := unix.Unmount(mp, mntDetach)
   if err != nil && err != unix.EINVAL {
      return &os.PathError{Op: "UnmountAll", Path: mp, Err: err}
   }
}
```

after which we can read mountinfo and see if there's anything left.

Now, again, this optimization only makes sense if we assume that in most of the cases the mounts will be a direct descendants of the `parent` (which is also true for your code).

PR #63, opened 2021-02-16
[WIP] ci/gha: enable doc generation

kolyshkin:
Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>


kolyshkin:
OK I spent some time and I still can't make it work :(

... and even if it will, the result is a zip archive (GHA currently does not support anything else), so it's not really easy to view html pages.

Closing

PR #65, opened 2021-02-17
ci/gha: add go 1.16; run as root; fix some tests; add fedora

kolyshkin:
A few tiny improvements to GHA CI:
 - add go 1.16 to testing matix :fire: ;
 - explicitly specify `ubuntu-20.04` (rather than `ubuntu-latest`);
 - move 'make lint' and 'make cross' to before 'make test';
 - update actions/setup-go to v2;
 - add :womans_hat: fedora  runner (to test functionality that needs `openat2(2)`);
 - remove remaining travis-ci bits;
 - fix mount tests for kernel 5.9+;
 - run tests as root :fire:. 

The last item revealed that tests that require `openat2` do not check for it :see_no_evil: :, so add the check.


kolyshkin:
@cpuguy83 @thaJeztah PTAL. No change in package functionality, just CI infra and a bit of tests.

thaJeztah (changes requested):

thaJeztah:
We should probably also switch to `actions/setup-go@v2`; https://github.com/actions/setup-go/tree/v2.1.3#v2

> The V2 offers:
> 
> - Adds GOBIN to the PATH
> - Proxy Support
> - stable input
> - Bug Fixes (including issues around version matching and semver)

cpuguy83:
CI shows passing but:

image

kolyshkin:
> CI shows passing but

This is OK, the kernel is not that new on GHA so some tests are skipped (unless inside a vagrant vm with fedora -- which is why I am adding it.

Now, I'm not sure why skipping is reported as error (I use `t.Skip()`)

.github/workflows/test.yml
```diff
@@ -14,10 +14,33 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
       uses: actions/checkout@v2
-    - name: Test
+    - name: Test (!linux)
+      if: ${{ runner.os != 'Linux' }}
       run: make test
+    - name: Test (linux)
+      if: ${{ runner.os == 'Linux' }}
+      run: sudo make test
```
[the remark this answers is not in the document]
kolyshkin: Yes (otherwise the tests will use stock go version, not the one we install above)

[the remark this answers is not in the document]
kolyshkin: fixed

kolyshkin:
> We should probably also switch to actions/setup-go@v2

added; PTAL @cpuguy83 @thaJeztah

.github/workflows/test.yml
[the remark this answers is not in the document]
kolyshkin: Note that it screws up some permissions (like one for `$HOME/.cache`) but I fix it with `chown` and making `sudo make test` the last command.

.github/workflows/test.yml
```diff
@@ -4,20 +4,46 @@ jobs:
...
       run: make cross
+    - name: Test (!linux)
+      if: ${{ runner.os != 'Linux' }}
+      run: make test
+    - name: Test (linux)
+      if: ${{ runner.os == 'Linux' }}
+      run: |
+        sudo -E PATH=$PATH make test
+        # fixup go cache screwed by sudo
+        sudo chown -R $(id -u):$(id -g) $HOME/.cache
```
thaJeztah, line 30: Some questions; 

- Is the cache actually used (or "useful"), assuming these nodes are cleaned between runs?
- Did this cause a problem in CI, or only on your local machine (testing `sudo -E PATH=... `, after which the files were owned by `root`)? (if so, it may not be needed in CI, as state would be cleared after this completes, correct?)
- Alternatively, we can set `GOCACHE` so that the `sudo` variant writes to it's own home directory, not the home directory of the non-privileged user, e.g.

    ```bash
    sudo -E PATH=$PATH GOCACHE=/root/.cache make test
    ```

thaJeztah:
> Update2: I think this is some unfortunate GHA cleverness (highlighting the function not implemented as an error or so)

Could it be it prepends `Error: ` for any line on `STDERR` ?

.ci/Vagrantfile.fedora33
```diff
@@ -20,8 +20,7 @@ Vagrant.configure("2") do |config|
       cat << EOF | dnf -y shell && break
 config exclude kernel,kernel-core
```
thaJeztah, line 21: Wondering if this line was related to the `update` line you removed (i.e.; "update, but don't install kernel updates")?

thaJeztah (commented):
Overall looks good (thanks!) left some questions

cpuguy83 (approved):
LGTM

.github/workflows/test.yml
kolyshkin: > Is the cache actually used (or "useful"), assuming these nodes are cleaned between runs?

It is used (based by the errors in CI I saw), and I guess it is mildly useful. Surely, we're not compiling docker here, so the savings are not that big.

I am fixing the .cache ownership not because the cache is useful, but for the sake of future contributors who might add some commands after that one, and discover that it fails.

Actually, the proper way would be to compile the tests as a user and run them as root, but I'm not sure how to do it in an elegant way.

> Did this cause a problem in CI

Yes it did, as "make lint" and "make cross" were before "sudo make test" and they failed because they could not write to `~/.cache`. So I mitigated it with two things -- (1) moved all non-sudo commands to before sudo ones (2) added this fixup.

> we can set `GOCACHE`

Good idea, let me try. Judging by the amount of code to compile (our packages plus a bits of x/sys/unix), caching should not result in any dramatic improvements.

kolyshkin: Ended up implementing decent `make sudotest` via Makefile, which compiles tests as a user and runs them as root. This way we don't have to care about gocache and other similar things (like re-downloading modules).

kolyshkin:
> Could it be it prepends Error: for any line on STDERR ?

I am using `t.Skipf`, and the docs say `Skipf is equivalent to Logf followed by SkipNow`. This means that either any `t.Log[f]` goes to stderr, or it doesn't, iow there's nothing special about message printed by `t.Skipf`.

So I stll think there's some special cleverness in GHA to highlight a line containing 'function not implemented'. The only way to check is to remove this message, I'll try that (but I don't want to remove message forever).

.ci/Vagrantfile.fedora33
kolyshkin: Yes it is, but I'd like to keep this line, which basically says "no matter what we do below, exclude kernel from it".

Some Fedora version needed update in the past, and this need might re-surface, so there's a chance someone will add `update` later, and they will definitely forget about excluding the kernel, which is a big package for which there's almost always an update available.

So yes, the line does not make sense per se currently, but it might come handy later, and since it does not add any overhead or something, I'd rather keep it.

Alternatively, we can re-add `update` and comment it out, together with the exclude. Let me know if you prefer it this way.

kolyshkin: (given that the comments are available in that context -- I'm not entirely sure)

kolyshkin:
Scratch that, I found a better way to run go test via sudo

.github/workflows/test.yml
kolyshkin: Scratch that bloody sudotest, I found a _much_ better way, see the last commit.

kolyshkin:
@thaJeztah PTAL

mountinfo/mounted_linux_test.go
```diff
@@ -174,10 +174,20 @@ func cleanupMounts(t *testing.T, dir string, mounts []string) {
...
 	}
+	fd, err := unix.Openat2(unix.AT_FDCWD, ".", &unix.OpenHow{Flags: unix.O_RDONLY})
+	if err != nil {
+		t.Skipf("openat2: %v (old kernel? need Linux 5.6+)", err)
+	}
+	unix.Close(fd)
+}
+
+func TestMountedBy(t *testing.T) {
+	requireOpenat2(t)
```
thaJeztah, line 190: Do you think it would make sense to only skip the parts that require openat2, instead of the whole test? https://github.com/moby/sys/pull/67

thaJeztah (approved):
LGTM ❤️

left one question (could do in a follow up)

AkihiroSuda (approved):

mountinfo/mounted_linux_test.go
kolyshkin: I definitely had openat2 way in mind when writing this, mostly added mountinfo to test my test :) 

But yeah, when I look at it, it makes some sense.

kolyshkin: But as we already test in on an openat2-capable system that should be sufficient.

IOW it's up to you whether you want to update #67 to run this test partially or just close it.

PR #67, opened 2021-03-04
TestMountedBy: skip mountedByOpenat2() check if not supported

thaJeztah:
This updates the test to only skip the mountedByOpenat2() part, instead
of skipping the whole test, if openat2 is not supported.

### DIFF
--- a/mountinfo/mounted_linux_test.go
+++ b/mountinfo/mounted_linux_test.go
@@ -179,22 +179,27 @@ func requireOpenat2(t *testing.T) {
 	if os.Getuid() != 0 {
 		t.Skip("requires root")
 	}
-	fd, err := unix.Openat2(unix.AT_FDCWD, ".", &unix.OpenHow{Flags: unix.O_RDONLY})
-	if err != nil {
+	if err := tryOpenat2(); err != nil {
 		t.Skipf("openat2: %v (old kernel? need Linux 5.6+)", err)
 	}
-	unix.Close(fd)
+}
+
+func tryOpenat2() error {
+	fd, err := unix.Openat2(unix.AT_FDCWD, ".", &unix.OpenHow{Flags: unix.O_RDONLY})
+	if err == nil {
+		_ = unix.Close(fd)
+	}
+	return err
 }
 
 func TestMountedBy(t *testing.T) {
-	requireOpenat2(t)
-
 	dir, mounts, err := prepareMounts(t)
 	defer cleanupMounts(t, dir, mounts)
 	if err != nil {
 		t.Fatalf("setup failed: %v", err)
 	}
 
+	openat2Supported := tryOpenat2() == nil
 	checked := false
 	for _, m := range mounts {
 		exp := !strings.Contains(m, notMounted)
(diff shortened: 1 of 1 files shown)


thaJeztah:
@kolyshkin PTAL; noticed this when running tests in a container on Docker Desktop (kernel 4.19)

thaJeztah:
ehm. probably don't need this as https://github.com/moby/sys/pull/65 also fixes this 🤦‍♂️thought we already merged that one

AkihiroSuda:
Merged https://github.com/moby/sys/pull/65

kolyshkin:
As I said in #65, I definitely had openat2 way in mind when writing this, mostly added mountinfo to test the test itself. As we already test this functionality on an openat2-capable system that should be sufficient, i.e. this check does not bring much by itself.

Feel free to either enable part of GetMountedBy test, or close this PR.

thaJeztah:
I updated the PR, but would be fine either way (closing / merging). ptal

mountinfo/mounted_linux_test.go
```diff
@@ -179,16 +179,21 @@ func requireOpenat2(t *testing.T) {
...
+	if err := tryOpenat2(t); err != nil {
 		t.Skipf("openat2: %v (old kernel? need Linux 5.6+)", err)
 	}
-	unix.Close(fd)
 }
 
-func TestMountedBy(t *testing.T) {
-	requireOpenat2(t)
+func tryOpenat2(t *testing.T) error {
+	t.Helper()
```
kolyshkin, line 188: AFAIK it does not have to be a helper (so also no need to pass `t`).

mountinfo/mounted_linux_test.go
```diff
@@ -207,11 +212,15 @@ func TestMountedBy(t *testing.T) {
 		}
 
 		checked = true
-		mounted, err = mountedByOpenat2(m)
-		if err != nil {
-			t.Errorf("mountedByOpenat2(%q) error: %v", m, err)
-		} else if mounted != exp {
-			t.Errorf("mountedByOpenat2(%q): expected %v, got %v", m, exp, mounted)
+
+		err = tryOpenat2(t)
```
kolyshkin, line 216: It doesn't matter much really, but calling it in a loop seems unclean -- e.g. it could make a reviewer think that something we do here might affect the value returned from `tryOpenat2`.

So I would move it out of the loop, doing something like
```go
openat2Supported = tryOpenat2(t) == nil
```
and below
```go
if openat2Supported {
   mounted, err = mountedByOpenat2(m)
   ...
```

kolyshkin:
> I updated the PR, but would be fine either way (closing / merging). ptal

Left some notes, also the commit description needs an update.

mountinfo/mounted_linux_test.go
kolyshkin: @thaJeztah PTAL ^^^

mountinfo/mounted_linux_test.go
thaJeztah: Yeah, probably not strictly needed; I usually add it by default, so that issues get reported with the line of the Test, not the line in the _helper_. I'll remove

thaJeztah:
@kolyshkin updated; ptal

kolyshkin:
LGTM except the commit message seems to be not up-to-date (it was before #65).

kolyshkin:
> LGTM except the commit message seems to be not up-to-date (it was before #65).

@thaJeztah PTAL ^^^ 🙏🏻

thaJeztah:
@kolyshkin updated, PTAL

kolyshkin (approved):
LGTM, thanks!

PR #68, opened 2021-03-09
mount: also test against local mountinfo source code

thaJeztah:
mount depends on mountinfo, but defaults to the release specified in go.mod.

This change adds a new test to also run against the local code so that we can catch regressions / breaking changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,14 +6,30 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
 SUDO ?= sudo -n
 
 .PHONY: all
-all: lint test cross
+all: clean lint test test-local cross
+
+.PHONY: clean
+clean:
+	$(RM) mount/go-local.*
 
 .PHONY: test
 test: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
-test:
+test: test-local
 	for p in $(PACKAGES); do \
 		(cd $$p && go test $(RUN_VIA_SUDO) -v .); \
 	done
+
+# test the mount module against the local mountinfo source code instead of the
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local:
+	echo 'replace github.com/moby/sys/mountinfo => ../mountinfo' | cat mount/go.mod - > mount/go-local.mod
+	cd mount \
+	&& go mod download -modfile=go-local.mod \
+	&& go test -modfile=go-local.mod $(RUN_VIA_SUDO) -v .
+	$(RM) mount/go-local.*
 
 .PHONY: lint
 lint: $(BINDIR)/golangci-lint
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/mount/go-local.*


Makefile
```diff
@@ -6,15 +6,31 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local: clean
+	echo "replace github.com/moby/sys/mountinfo => ../mountinfo" | cat mount/go.mod - > mount/go-local.mod \
+	&& cd mount \
+	&& go test -modfile=go-local.mod $(RUN_VIA_SUDO) -v . ;\
+	rm go-local.*
+	cd ../
```
thaJeztah, line 32: I must admit that I've been going back-and-forth what the cleanest way is to implement this;

- use `mktemp -d` to create a temporary directory for the `go-local.mod` ?
- note that bash "pseudo files" (`<(echo "replace github.com/moby/sys/mountinfo => ../mountinfo" | cat mount/go.mod -)`) don't work; `go` _REQUIRES_ a file, and it must have a `.mod` extension (possibly something we should request to be changed, so that, e.g. `STDIN` / `-` could be supported?
- combine this with the `test:` target? (less flexible, but perhaps cleaner?)
- split creation of `mount/go-local.md` to a separate target?

Input/suggestions welcome; my initial goal here is to have _some_ way to test against the local source 😅

thaJeztah:
@kolyshkin ptal; happy to hear your thoughts on this

kolyshkin:
Perhaps we can just add the `replace` line to go.mod, do the test and remove it. Add it with a comment to have a marker to aid in removal.

Something like this (as a script in mount directory, UNTESTED):

```shell
#!/bin/bash
MARKER='# Added temporarily by test-latest'

trap "sed -i \"/^${MARKER}/,$d\" go.mod" EXIT

if ! grep -q "^${MARKER}$" go.mod; then
        cat << EOF >> go.mod
$MARKER
replace replace github.com/moby/sys/mountinfo => ../mountinfo
EOF
fi

go test -v .
```

And in the top makefile:

```makefile
.PHONY: test-latest
test-latest:
        ./mount/test-latest
```

Can also be done entirely in Makefile, of course...

thaJeztah:
> Perhaps we can just add the replace line to go.mod, do the test and remove it. Add it with a comment to have a marker to aid in removal.

I considered something like that, but (personally) prefer not to modify files that are in version control; that way we also don't have to revert the change (and allow running it locally as well, without accidentally modifying the file)

.github/workflows/test.yml
```diff
@@ -39,4 +41,4 @@ jobs:
         run: ssh default 'sh -exc "uname -a && df -T"'
 
       - name: tests
-        run: ssh default 'cd /vagrant && make test'
+        run: ssh default 'cd /vagrant && make test && make test-local'
```
kolyshkin, line 44: can be just `make test test-local`

Makefile
```diff
@@ -6,15 +6,32 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
+# test the mount module against the local mountinfo source code instead of the
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local: clean
+	echo 'replace github.com/moby/sys/mountinfo => ../mountinfo' | cat mount/go.mod - > mount/go-local.mod \
+	&& cd mount \
+	&& go mod download -modfile=go-local.mod \
+	&& go test -modfile=go-local.mod $(RUN_VIA_SUDO) -v . ;\
```
kolyshkin, line 31: nit: add a space before `\`.

Makefile
```diff
@@ -6,15 +6,32 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
 		(cd $$p && go test $(RUN_VIA_SUDO) -v .); \
 	done
 
+# test the mount module against the local mountinfo source code instead of the
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local: clean
+	echo 'replace github.com/moby/sys/mountinfo => ../mountinfo' | cat mount/go.mod - > mount/go-local.mod \
```
kolyshkin, line 28: Here continuation (`\` and `&&` on the next line) is not needed. Make executes the commands via shell one by one, if a command fails, it stops.

The continuation is needed after `cd` as it won't affect the next command.

Makefile
```diff
@@ -6,15 +6,32 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
+	&& cd mount \
+	&& go mod download -modfile=go-local.mod \
+	&& go test -modfile=go-local.mod $(RUN_VIA_SUDO) -v . ;\
+	rm go-local.*
+	cd ../
```
kolyshkin, line 33: This line is not needed (every command is run via its own shell).

Makefile
```diff
@@ -6,15 +6,32 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
+all: clean lint test test-local cross
+
+.PHONY: clean
+clean:
+	$(RM) mount/go-local.*
 
 .PHONY: test
 test: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
-test:
+test: clean
```
kolyshkin, line 17: Not needed? Even if `mount/go-local.mod` is present, it does not affect `go test` in any way (AFAIK).

.github/workflows/test.yml
kolyshkin: ...but rather than having a need to call `test-local` separately, I'd add `test: test-local` dependency into the Makefile.

Makefile
```diff
@@ -6,15 +6,32 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
 	for p in $(PACKAGES); do \
 		(cd $$p && go test $(RUN_VIA_SUDO) -v .); \
 	done
 
+# test the mount module against the local mountinfo source code instead of the
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local: clean
```
kolyshkin, line 27: `clean` dependency is not needed, as we're overwriting the file anyway.

Makefile
thaJeztah: Ah! Good one; I sometimes get confused when to `\` and when not in Makefiles

Makefile
thaJeztah: With your other comments about continuation; I'll remove this, and make the rm just `rm -f mount/go-local.*`

.github/workflows/test.yml
thaJeztah: Yes, makes sense; tests run really fast, so shouldn't be an issue to always run both

kolyshkin:
@thaJeztah PTAL

Makefile
```diff
@@ -6,15 +6,31 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
 .PHONY: all
-all: lint test cross
+all: clean lint test test-local cross
+
+.PHONY: clean
+clean:
+	$(RM) mount/go-local.*
 
 .PHONY: test
 test: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
```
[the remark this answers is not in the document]
thaJeztah: > (and move up) to avoid repetition.

You mean, move `test-local` above `test` ?

Makefile
```diff
@@ -6,15 +6,31 @@ CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
...
+# test the mount module against the local mountinfo source code instead of the
+# release specified in go.mod. This allows catching regressions / breaking
+# changes in mountinfo.
+.PHONY: test-local
+test-local: RUN_VIA_SUDO = $(shell $(SUDO) true && echo -exec \"$(SUDO)\")
+test-local:
+	echo 'replace github.com/moby/sys/mountinfo => ../mountinfo' | cat mount/go.mod - > mount/go-local.mod
+	cd mount \
+	&& go mod download -modfile=go-local.mod \
+	&& go test -modfile=go-local.mod $(RUN_VIA_SUDO) -v .
```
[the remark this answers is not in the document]
thaJeztah: I'm not sure if understand this change; `go-local.sum` will be _different_ than `go.sum`, because it's using different versions of the dependency; the `go mod download -modfile=go-local.mod` may pull other versions of dependencies, because `mountinfo` ("main branch") may be using different versions of dependencies.

PR #76, opened 2021-08-07
mountinfo.GetMountsFromReader() use strings.SplitN()

thaJeztah:
We expect 2 fields to be there, anything below (or above) will produce
an error, so limit the number of strings we split into to 3. This still
allows us to detect "too many colons", but won't result in splitting
without bounds.


kolyshkin (approved):
LGTM

thaJeztah:
@AkihiroSuda ptal

AkihiroSuda (approved):

PR #77, opened 2021-08-07
signal: remove DefaultStopSignal

thaJeztah:
relates to https://github.com/moby/moby/pull/42717

This const was inherited from the Docker/Moby repository, but does not make a lot of sense in the new location.

Removing this const here.


samuelkarp (approved):

tianon (approved):
LGTM

PR #79, opened 2021-08-13
lint: update golangci-lint to v1.41.1

thaJeztah:


tianon (approved):

kolyshkin (approved):
LGTM

Ideally, I think, we should drop installing aux tools from the Makefile, and use golangci-lint action job in GHA CI

kolyshkin:
... and also enable dependabot (which would help in case of GHA CI job, but not in this particular case)

PR #81, opened 2021-08-14
mountinfo: GetMountsFromReader() remove redundant switch, fix dropped "optional" fields

thaJeztah:
This switch was implementing the exact same check as is done in
strings.Join(); https://cs.opensource.google/go/go/+/refs/tags/go1.16.7:src/strings/strings.go;l=421-427

```go
func Join(elems []string, sep string) string {
	switch len(elems) {
	case 0:
		return ""
	case 1:
		return elems[0]
	}
```

This only difference removing this switch makes is that it will assign
an empty string to p.Optional if no optional fields are present, but
given that an empty string is the default value for p.Optional, this
should make no difference.

This also revealed an off-by-one bug:

Before this change:

- we return an error if `sepIdx == 5` (earlier check)
- for `sepIdx == 6` we skipped (in this switch)
- for `7`, we take field 6 (equivalent to `fields[6:7]`, or `fields[6:sepIdx]`)
- for the "default" (else), we use `fields[6:sepIdx-1]`, which for (e.g.) `sepIndex=8`
  would do the same as `sepIdx=7`, so we'd drop one value.

### DIFF
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -105,16 +105,9 @@ func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
 		p.Options = fields[5]
 
 		// zero or more optional fields
-		switch {
-		case sepIdx == 6:
-			// zero, do nothing
-		case sepIdx == 7:
-			p.Optional = fields[6]
-		default:
-			p.Optional = strings.Join(fields[6:sepIdx-1], " ")
-		}
+		p.Optional = strings.Join(fields[6:sepIdx], " ")
 
-		// Run the filter after parsing all of the fields.
+		// Run the filter after parsing all fields.
 		var skip, stop bool
 		if filter != nil {
 			skip, stop = filter(p)
--- a/mountinfo/mountinfo_linux_test.go
+++ b/mountinfo/mountinfo_linux_test.go
@@ -676,7 +676,7 @@ func TestParseMountinfoExtraCases(t *testing.T) {
 			name:  "one optional field",
 			entry: `251 15 0:3573 / /mnt/point rw,relatime shared:123 - aufs none rw`,
 			valid: true,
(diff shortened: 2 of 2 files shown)


mountinfo/mountinfo_linux.go
```diff
@@ -105,16 +105,9 @@ func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
 		p.Options = fields[5]
 
 		// zero or more optional fields
-		switch {
-		case sepIdx == 6:
-			// zero, do nothing
-		case sepIdx == 7:
-			p.Optional = fields[6]
-		default:
-			p.Optional = strings.Join(fields[6:sepIdx-1], " ")
```
thaJeztah, line 114: Looks like there's an off-by-one bug here; removing this switch caused it to panic;

```
--- FAIL: TestMountedBy (0.01s)
panic: runtime error: slice bounds out of range [6:5] [recovered]
	panic: runtime error: slice bounds out of range [6:5]
```

- we return an error if `sepIdx == 5` (earlier check)
- for `sepIdx == 6` we skipped (in this switch)
- for `7`, we take field 6 (equivalent to `fields[6:7]`, or `fields[6:sepIdx]`)
- for the "default" (else), we use `fields[6:sepIdx-1]`, which for (e.g.) `sepIndex=8` would do the same as `sepIdx=7`, so we'd drop one value

thaJeztah:
@kolyshkin PTAL

thaJeztah:
Looking further, I noticed that TestParseMountinfoExtraCases was not checking the `Optional` field; fixing that test showed that it was indeed broken without this patch, so I included those commits in this PR

kolyshkin (approved):
LGTM

thaJeztah:
@AkihiroSuda ptal

AkihiroSuda (approved):

PR #82, opened 2021-08-14
mountinfo: rewrite more tests to use a test-table

thaJeztah:
See individual commits for details


kolyshkin:
To me, the first commit definitely make sense (and is something I wanted to do myself but never got around to), the second one kind of make sense (if we have the data, why not use it for testing)?

The last commit seems useless. It makes sense in runc, as it was added together with a specific fix, to make sure runc parser can parse multiple cgroup v1 mount hierarchies (rather than one which is usual). Here it doesn't make sense since the parser is generic and doesn't care about how many cgroup v1 mount hierarchies are there.

Can we drop the last commit?

thaJeztah:
> The last commit seems useless. It makes sense in runc, as it was added together with a specific fix, to make sure runc parser can parse multiple cgroup v1 mount hierarchies (rather than one which is usual). Here it doesn't make sense since the parser is generic and doesn't care about how many cgroup v1 mount hierarchies are there.

Ah, yes. I initially was looking if it would make sense to have (some of) the cgroups helper in this repository as well (and for that reason, to have test coverage), but currently that's not there, so they don't add much. I'll drop that commit.

thaJeztah:
Updated to drop the last commit 👍

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

PR #83, opened 2021-08-14
mountinfo: GetMountsFromReader() remove workaround for kernel < 3.10

thaJeztah:
follow-up to https://github.com/moby/sys/pull/81

In Linux <= 3.9 mounting a cifs with spaces in a share name (like "//srv/My Docs")
_may_ end up having a space in the last field of mountinfo (like "unc=//serv/My Docs").

Since kernel 3.10-rc1, cifs option "unc=" is ignored, so spaces should not appear.

Given that all major distros are now on at least kernel 3.10, and (e.g.) docker
has not supported older kernel versions for 5 Years [1], it should be save to
remove this special handling.

[1]: https://github.com/moby/moby/commit/51b23d88422918c24291f5876df35f91b23a446a


mountinfo/mountinfo_linux_test.go
```diff
@@ -504,66 +499,6 @@ func TestParseFedoraMountinfoFilterFields(t *testing.T) {
 	}
 }
 
-func TestParseMountinfoWithSpaces(t *testing.T) {
```
kolyshkin, line 507: I don't understand how removing this test is relevant to this commit.

kolyshkin:
We should probably do this once Go < 1.23 (which raises the Linux kernel requirement to >= 3.16 or so) is no longer supported.

mountinfo/mountinfo_linux_test.go
kolyshkin: To reiterate the comment above -- this test case is NOT about pre-3.10 bug of having extra spaces in a cifs mount entry where share name has spaces.

kolyshkin: @thaJeztah ^^^ (or do we have other tests where we test fields with spaces)?

thaJeztah: >  this test case is NOT about pre-3.10 bug of having extra spaces in a cifs mount entry where share name has spaces.

Oh! Right, only the second entry is about the `unc=<path with spaces>` issue; the others are escaped

IIUC from the comment, kernel 3.10-rc1 should ignore that `unc` option, so I can just remove that part.

Do we know of other ways for spaces to get into the last column, or should  3.10-rc1 and up guarantee there not to be?

> (or do we have other tests where we test fields with spaces)?

@kolyshkin hm.. not sure if we have; I'll add back this test, but update it to only remove the `unc=` bits.

thaJeztah:
> We should probably do this once Go < 1.23 (which raises the Linux kernel requirement to >= 3.16 or so) is no longer supported.


Slightly wondering if we need to wait for that; Kernel 3.9 (or older) means;

- Debian 7 Wheezy (kernel 3.2); EOL since 2018
- Ubuntu 13.04 (kernel 3.8); EOL since 2014
- CentOS 6 / RHEL 6 / Fedora 10 (kernel 2.6)

kolyshkin:
> We should probably do this once Go < 1.23 (which raises the Linux kernel requirement to >= 3.16 or so) is no longer supported.

To correct myself: Go 1.24 (now released) requires Linux 3.2 or later. Once we'll switch to Go version which will require kernel 3.10 or later, this can and should be removed. Can do it earlier but I see no big reason why. I'd rather be conservative here and keep this code for a while (especially since it is neither complex nor creates any noticeable performance overhead for a default scenario).

PR #84, opened 2021-09-20
ci/gha: drop go 1.14, add go 1.17

kolyshkin:
go 1.15 is no longer supported, and go 1.14 is not supported since go
1.16 is released (roughly half a year ago).

Let's keep the latest unsupported version (1.15) for now, but drop the
older 1.14.


AkihiroSuda (approved):

thaJeztah (approved):
LGTM

PR #85, opened 2021-10-09
mountinfo: support darwin build

thehajime:
This PR allows us to use mountinfo by darwin programs, which is used in this PR (via mount.Lookup() of containerd).

https://github.com/containerd/containerd/pull/5935


kolyshkin (approved):
LGTM assuming CI is 🍏

AkihiroSuda (approved):

Issue #88, opened 2021-10-25
mountinfo.Mounted: incorrect results for some corner cases


kolyshkin:
As reported by @mkimuram here: https://github.com/kubernetes/kubernetes/pull/105833#issuecomment-951092014

Obviously, issues a and b can be fixed by using `filepath.EvalSymlinks` (and in view of that part of https://github.com/moby/sys/pull/86 might need to be reverted).

Not sure what to do with issue c for now...

PR #90, opened 2021-10-28
mountinfo.Mounted(): fix

kolyshkin:
Fix Mounted() and its tests.

1. In case the argument does not exist, Mounted now returns an
       appropriate error. Previously, this condition was treated as
       "not a mount point", with no error returned, but in fact such
       path could be a mount point, overshadowed by another mount.
       Since the mount is still unreachable, it can not be unmounted
       anyway, so we do not fall back to `mountedByMountinfo`, returning
       an error instead.
    
     ⚠️ In case callers are not interested in `ENOENT`, they should filter it
     out. Amend the function doc accordingly.
    
2. There are some corner cases in which `mountedByOpenat2` or `mountedByStat`
       do not work properly. In particular,
    
    - neither work if there's an ending slash;
    - neither work for a symlink to a mounted directory;
    - `mountedByStat` do not work a parent of path is a symlink residing on
      different filesystem.
    
   To solve these cases, call `normalizePath()` before path is used for any of
   `mountedBy*` functions, not just before `mountedByMountinfo`
   like it was done earlier.

3. Fix `TestMountedBy`

    The way `prepareMounts` was written was very complicated from the
    maintainability perspective, with all the test cases in the same basket.
    
    Rewrite this in a much cleaner way, with individual subtests.
    
    Add some more test cases that were failing before this PR.

Fixes: #88

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -16,9 +16,6 @@ func mountedByOpenat2(path string) (bool, error) {
 		Flags: unix.O_PATH | unix.O_CLOEXEC,
 	})
 	if err != nil {
-		if err == unix.ENOENT { // not a mount
-			return false, nil
-		}
 		return false, &os.PathError{Op: "openat2", Path: dir, Err: err}
 	}
 	fd, err := unix.Openat2(dirfd, last, &unix.OpenHow{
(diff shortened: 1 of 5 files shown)


thaJeztah:
CI on my test PR (https://github.com/moby/moby/pull/42980) didn't explode, so LGTM

@mkimuram @kolyshkin are any changes still needed, for the discussion on https://github.com/moby/sys/pull/90#discussion_r738646397 above, or is this good to go?

mkimuram:
Sorry, I should have clearly commented that it is good to go. It looks good to me.

thaJeztah:
Thanks! Just double-checking 😅 

Thanks for helping review these changes; it's always good to have more eyes 🤗

thaJeztah (approved):
LGTM

PR #95, opened 2021-12-03
mountinfo: introduce MountedBind

kolyshkin:
`MountedBind` is a Linux-specific function, similar to Mounted, but with
additional workarounds to decrease the possibility of misdetecting
a bind mount.

On Linux (before kernel 5.8, which includes [commit 9f6c61f96f2d97](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9f6c61f96f2d97cbb5f7fa85607bc398f843ff0f), and when
`openat2()` syscall (appeared in kernel 5.6) is not available) bind mount detection
is not reliable (false negatives are possible) on systems that perform
lots of mounts and unmounts, due to a kernel issue.

This function tries to mitigate that issue by re-reading mountinfo, thus
decreasing the possibility of a false negative.

This function increases system load. It should only be used for bind
mounts, if a user wants to sacrifice system performance for higher reliability
to detect a bind mount.

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -7,6 +7,82 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// MountedBind is a Linux-specific function, similar to Mounted, but with
+// additional workarounds to decrease the possibility of misdetecting
+// a bind mount.
+//
+// On Linux (before kernel 5.8) bind mount detection is not reliable (false
+// negatives are possible) on systems that perform lots of mounts and unmounts,
+// due to a kernel issue. This function tries to mitigate that issue by
+// re-reading mountinfo, decreasing the possibility of a false negative.
+//
+// This function increases system load. It should only be used for bind
+// mounts, if a user wants to sacrifice system performance for higher reliability.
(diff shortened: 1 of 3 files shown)


mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,82 @@ import (
...
+		}
+		total++
+		return true, false // skip, keep going
+	})
+	if err != nil {
+		return false, err
+	}
+	if len(entries) > 0 {
+		return true, nil
+	}
```
manugupt1, line 70: Would it not return a true without a retry if a mount point is found?

kolyshkin: Yes, the re-read is only for negative cases. Once the mount point is found, there's absolutely need for any kind of retries or re-reads (note it is different from kubernetes approach, which _always_ reads mountinfo at least twice).

thaJeztah:
@kolyshkin this one still "DRAFT"? (I see you removed `WIP` from the title)

kolyshkin:
Closing in favor of #100

PR #96, opened 2021-12-04
Expose MountedByOpenat2 as a public function so that k8s can consume

manugupt1:
Expose MountedByOpenat2 as a public function so that k8s can consume

One of the alternatives suggested here is to expose MountedByOpenat2 as a public function
so that k8s can consume and rest of the downstream can also benefit from it.

https://github.com/kubernetes/kubernetes/pull/105833#issuecomment-985591009

cc @kolyshkin

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -7,9 +7,9 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// mountedByOpenat2 is a method of detecting a mount that works for all kinds
+// MountedByOpenat2 is a method of detecting a mount that works for all kinds
 // of mounts (incl. bind mounts), but requires a recent (v5.6+) linux kernel.
-func mountedByOpenat2(path string) (bool, error) {
+func MountedByOpenat2(path string) (bool, error) {
 	dir, last := filepath.Split(path)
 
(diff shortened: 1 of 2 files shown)


kolyshkin:
I think it's better to export something like 'MountedFast' which includes all the methods except for parsing mountinfo.

I'll create the PR.

kolyshkin:
> I think it's better to export something like 'MountedFast' which includes all the methods except for parsing mountinfo.

OK, this is problematic, because there's no way to communicate back to the caller if we're sure this is not a mount point.

Unless we do something like this:

```go
func MountedFast(path string) (mounted, sure bool, err error)
```

when `mounted == false`, the caller needs to check `sure` and if it's false, proceed to mountinfo parsing.

manugupt1:
> > I think it's better to export something like 'MountedFast' which includes all the methods except for parsing mountinfo.
> 
> OK, this is problematic, because there's no way to communicate back to the caller if we're sure this is not a mount point.
> 
> Unless we do something like this:
> 
> ```go
> func MountedFast(path string) (mounted, sure bool, err error)
> ```

That is something that will work to; my PR always falls back if the fast path cannot guarantee a mount-point.

manugupt1:
Closing in favor of https://github.com/moby/sys/pull/97

PR #97, opened 2021-12-08
Introduce MountedFast function to determine mount points faster.

manugupt1:
MountedFast is a method of detecting mount points really fast.
It works for all kinds of mounts (incl. bind mounts) but requires
a recent (v5.6+) Linux kernel. A mount-point check is guaranteed to
be a mount-point or not only when sure is true. When sure is false,
the caller needs to check for other methods (eg: parse /proc/mounts)
to successfully determine if it is a mount-point.

This has been motivated by
https://github.com/kubernetes/kubernetes/pull/105833#issuecomment-985591009
and
https://github.com/moby/sys/pull/96#issuecomment-985952881

## Context
The context is that; k8s relies on reading /proc/mounts to determine mount points. It also enforces a consistency check (ie ensure that two reads of proc/mounts are exactly the same) before determining a mount-point which leads to resource contention. This is different than docker where a lock is acquired before reading the entire /proc/mounts.

Now, k8s wont remove the consistency check as it could lead to all sorts of error that Kir demonstrated here (https://github.com/kolyshkin/procfs-test) and the consistency check mitigates that.

To solve for that k8s took inspiration from moby/sys to do its own openat2 call (https://github.com/kubernetes/kubernetes/pull/105833) and when openat2 is not available, fallback on reading /proc/mounts with the consistency check I mentioned above.

During the course of the discussion, Kir suggested it might be better to upstream the change to moby/sys and k8s can take it as a dependency which is what this PR does. It introduces a MountedFast  which uses openat2 (for kernel 5.6+) or stat calls to determine a mount-point, it sends back a sure bool to the caller to guarantee the caller if it is a mount-point or not.
k8s will now first call MountedFast when this change is merged and then if sure is false or an error is thrown, it will fall back to reading /proc/mounts. This is beneficial for k8s on newer kernels as they will never read /proc/mounts because of openat2 path and for older kernels only bind mounts will be validated using /proc/mounts.

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -7,6 +7,50 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// MountedFast is a method of detecting mount points without reading
+// mountinfo from procfs. A mount point check is guaranteed to be a
+// mount point or not only when sure is true. When sure is false, the
+// caller needs to check for other methods (eg: parse /proc/mounts)
+// to successfully determine if it is a mount point. This function
+// is only available on Linux.
+func MountedFast(path string) (mounted, sure bool, err error) {
+	// root is always mounted
+	if path == string(os.PathSeparator) {
+		return true, true, nil
+	}
+
+	path, err = normalizePath(path)
+	if err != nil {
+		return false, false, err
+	}
+
+	return mountedFast(path)
+}
+
+// mountedFast combines mountedByOpenAt2 and mountedByStat.
+// A mount point check is guaranteed to be a mount point or not only
+// when sure is true. When sure is false, the caller needs to check
+// for other methods (eg: parse /proc/mounts) to successfully
+// determine if it is a mount point.
+func mountedFast(normalizedPath string) (mounted, sure bool, err error) {
+	// Try a fast path, using openat2() with RESOLVE_NO_XDEV.
+	mounted, err = mountedByOpenat2(normalizedPath)
+	if err == nil {
+		return mounted, true, nil
+	}
+
+	// Another fast path: compare st.st_dev fields.
+	mounted, err = mountedByStat(normalizedPath)
+	// This does not work for bind mounts, so false negative
+	// is possible, therefore only trust if return is true.
+	if mounted && err == nil {
+		return mounted, true, nil
+	}
+
+	// NB: not sure
+	return false, false, err
+}
+
 // mountedByOpenat2 is a method of detecting a mount that works for all kinds
 // of mounts (incl. bind mounts), but requires a recent (v5.6+) linux kernel.
 func mountedByOpenat2(path string) (bool, error) {
@@ -39,16 +83,10 @@ func mounted(path string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	// Try a fast path, using openat2() with RESOLVE_NO_XDEV.
-	mounted, err := mountedByOpenat2(path)
-	if err == nil {
-		return mounted, nil
-	}
-	// Another fast path: compare st.st_dev fields.
-	mounted, err = mountedByStat(path)
-	// This does not work for bind mounts, so false negative
-	// is possible, therefore only trust if return is true.
-	if mounted && err == nil {
+
+	// Try all fast paths.
+	mounted, sure, err := mountedFast(path)
+	if sure && err == nil {
 		return mounted, nil
 	}
 
--- a/mountinfo/mounted_linux_test.go
+++ b/mountinfo/mounted_linux_test.go
@@ -5,6 +5,9 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
+	"runtime"
+	"strings"
 	"testing"
 
 	"golang.org/x/sys/unix"
@@ -276,8 +279,14 @@ func tryOpenat2() error {
 }
 
 func TestMountedBy(t *testing.T) {
-	openat2Supported := tryOpenat2() == nil
 	checked := false
+
+	// List of individual implementations to check.
+	toCheck := []func(string) (bool, error){mountedByMountinfo, mountedByStat}
+	if tryOpenat2() == nil {
+		toCheck = append(toCheck, mountedByOpenat2)
+	}
+
 	for _, tc := range testMounts {
 		tc := tc
 		t.Run(tc.desc, func(t *testing.T) {
@@ -316,45 +325,29 @@ func TestMountedBy(t *testing.T) {
 				t.Fatalf("normalizePath: %v", err)
 			}
 
-			mounted, err = mountedByMountinfo(m)
-			if err != nil {
-				t.Errorf("mountedByMountinfo error: %v", err)
-				// Check false is returned in error case.
-				if mounted != false {
-					t.Errorf("MountedByMountinfo: expected false on error, got %v", mounted)
-				}
-			} else if mounted != exp {
-				t.Errorf("mountedByMountinfo: expected %v, got %v", exp, mounted)
-			}
-			checked = true
+			for _, fn := range toCheck {
+				// Figure out function name.
+				name := runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 
-			mounted, err = mountedByStat(m)
-			if err != nil {
-				t.Errorf("mountedByStat error: %v", err)
-				// Check false is returned in error case.
-				if mounted != false {
-					t.Errorf("MountedByStat: expected false on error, got %v", mounted)
+				mounted, err = fn(m)
+				if err != nil {
+					t.Errorf("%s: %v", name, err)
+					// Check false is returned in error case.
+					if mounted != false {
+						t.Errorf("%s: expected false on error, got %v", name, mounted)
+					}
+				} else if mounted != exp {
+					if tc.isBind && strings.HasSuffix(name, "mountedByStat") {
+						// mountedByStat can not detect bind mounts.
+					} else {
+						t.Errorf("%s: expected %v, got %v", name, exp, mounted)
+					}
 				}
-			} else if mounted != exp && !tc.isBind { // mountedByStat can not detect bind mounts
-				t.Errorf("mountedByStat: expected %v, got %v", exp, mounted)
-			}
-
-			if !openat2Supported {
-				return
-			}
-			mounted, err = mountedByOpenat2(m)
-			if err != nil {
-				t.Errorf("mountedByOpenat2 error: %v", err)
-				// Check false is returned in error case.
-				if mounted != false {
-					t.Errorf("MountedByOpenat2: expected false on error, got %v", mounted)
-				}
-			} else if mounted != exp {
-				t.Errorf("mountedByOpenat2: expected %v, got %v", exp, mounted)
+				checked = true
 			}
 		})
-
 	}
+
 	if !checked {
 		t.Skip("no mounts to check")
 	}
--- a/mountinfo/mounted_linux_test.go
+++ b/mountinfo/mounted_linux_test.go
@@ -280,6 +280,7 @@ func tryOpenat2() error {
 
 func TestMountedBy(t *testing.T) {
 	checked := false
+	openat2Supported := tryOpenat2() == nil
 
 	// List of individual implementations to check.
 	toCheck := []func(string) (bool, error){mountedByMountinfo, mountedByStat}
@@ -311,6 +312,44 @@ func TestMountedBy(t *testing.T) {
 				// Check false is returned in error case.
 				if mounted != false {
 					t.Errorf("Mounted: expected false on error, got %v", mounted)
+				}
+			}
+
+			// Check the public MountedFast() function as a whole.
+			mounted, sure, err := MountedFast(m)
+
+			// For any error, regardless of kernel version, sure and mounted are false.
+			if err != nil {
+				if sure != false {
+					t.Errorf("MountFast: expected sure to be false, got %v", sure)
+				}
+				if mounted != false {
+					t.Errorf("MountFast: expected mounted as false on error, got %v", mounted)
+				}
+			} else {
+				if openat2Supported {
+					if mounted != exp {
+						t.Errorf("MountFast: expected mounted as %v, got %v", exp, mounted)
+					}
+				} else {
+					if tc.isBind {
+						if sure != false {
+							t.Errorf("MountFast: expected sure to be false, got %v", sure)
+						}
+						if mounted != false {
+							t.Errorf("MountFast: expected mounted to be false, got %v", mounted)
+						}
+					} else {
+						if mounted != exp {
+							t.Errorf("MountFast: expected mounted as %v, got %v", exp, mounted)
+						}
+						if tc.isMount && sure == false {
+							t.Errorf("MountFast: expected sure to be true, got %v", sure)
+						}
+						if !tc.isMount && sure == true {
+							t.Errorf("MountFast: expected sure to be false, got %v", sure)
+						}
+					}
 				}
 			}
 


manugupt1:
cc @kolyshkin  please have a look

mountinfo/mounted_linux.go
```diff
@@ -7,16 +7,40 @@ import (
...
+}
+
 // mountedByOpenat2 is a method of detecting a mount that works for all kinds
 // of mounts (incl. bind mounts), but requires a recent (v5.6+) linux kernel.
-func mountedByOpenat2(path string) (bool, error) {
+// A mount-point check is guaranteed to be a mount-point or not only when 
+// sure is true. When sure is false, the caller needs to check for other '
+// methods (eg: parse /proc/mounts) to successfully determine if it is a
+// mount-point.
+func mountedByOpenat2(path string) (mounted, sure bool, err error) {
```
kzys, line 36: How about returning an exported error instead of the sure flag?

Apparently in all error cases, sure=false. So returning an error wouldn't override the original errors.

Generally speaking Go code checks `err` first and doesn't consume other returning values `if err != nil`. I want to follow the convention here.

mountinfo/mounted_linux.go
```diff
@@ -40,8 +64,8 @@ func mounted(path string) (bool, error) {
 		return false, err
 	}
 	// Try a fast path, using openat2() with RESOLVE_NO_XDEV.
-	mounted, err := mountedByOpenat2(path)
-	if err == nil {
+	mounted, sure, err := mountedByOpenat2(path)
+	if mounted && sure && err == nil {
 		return mounted, nil
 	}
 	// Another fast path: compare st.st_dev fields.
```
kzys, line 71: 1. mountedByStat could be called from MountedFast
2. Moreover, how about calling MountedFast from Mounted?

mountinfo/mounted_linux.go
manugupt1: @kolyshkin Do you have thoughts as you suggested this? I think it boils down to preference.

mountinfo/mounted_linux.go
manugupt1: 1. Sure! Let me make that change.
2. We will be calling `normalizePath` twice for every `mounted` call. We can avoid this by relying on Os build tags and have `Mounted` for different OS. would you like me to make that refactor as well?

kzys: Ah sorry, I meant, Linux-specific `mounted` could call `MountedFast`. Regarding normalizePath, I honestly okay to call that multiple times. It wouldn't be that slow. Would it be?

mountinfo/mounted_linux.go
kolyshkin: I agree with @kzys here -- no need to change mountedByOpenat2 signature.

Basically, you need to take `mounted` from `mountinfo/mounted_linux.go`, add `sure bool` to it, and remove the fallback to mountinfo parsing. That's all what's needed.

kolyshkin:
Also, please cherry-pick https://github.com/moby/sys/pull/95/commits/03cca996870eabc66c0794bec30eb8abe2a29ffd since you're adding more implementations to test, and this becomes ugly -- and this commit is a nice solution to that.

mountinfo/mounted_linux.go
kolyshkin: `mounted` from `mountinfo/mounted_linux.go` already have all this, What's needed is to remove the fallback, instead returning `sure == false`.

mountinfo/mounted_linux.go
```diff
@@ -7,16 +7,39 @@ import (
...
+// MountedFast is a method of detecting mount points really fast.
+// A mount-point check is guaranteed to be a mount-point or not only
+// when sure is true. When sure is false, the caller needs to check
+// for other methods (eg: parse /proc/mounts) to successfully
+// determine if it is a mount-point.
+func MountedFast(path string) (mounted, sure bool, err error) {
+	// root is always mounted
+	if path == string(os.PathSeparator) {
+		return true, true, nil
+	}
```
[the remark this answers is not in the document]
manugupt1: I prefer having it here as it is a quick check and does not require use of syscalls. It is also faster.
The other thing is that os.PathSeparator is common between different OS's see: https://github.com/moby/sys/blob/master/mountinfo/mountinfo.go#L20-L23

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,50 @@ import (
...
+
+	return mountedFast(path)
+}
+
+// mountedFast combines mountedByOpenAt2 and mountedByStat
+// A mount-point check is guaranteed to be a mount-point or not only
+// when sure is true. When sure is false, the caller needs to check
+// for other methods (eg: parse /proc/mounts) to successfully
+// determine if it is a mount-point.
+func mountedFast(normalizedPath string) (mounted, sure bool, err error) {
```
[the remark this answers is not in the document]
manugupt1: Good call, done! thanks for catching that.

kzys (approved):

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,50 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// MountedFast is a method of detecting mount points without reading
+// mountinfo from procfs. A mount-point check is guaranteed to be a
+// mount-point or not only when sure is true. When sure is false, the
+// caller needs to check for other methods (eg: parse /proc/mounts)
+// to successfully determine if it is a mount-point.
+func MountedFast(path string) (mounted, sure bool, err error) {
```
jingxu97, line 15: basically we need three possible state, true, false, and not sure.
does use 1, 0, -1 or for return value a good choice or something like https://stackoverflow.com/questions/314800/best-way-to-define-true-false-unset-state?

jingxu97: or if not sure, we can return error indicating cannot figure out whether it is mounted or not. \

manugupt1: Hi, Jing,

I am fine with both options as long as @kolyshkin  does not have a preference. Would you be able to suggest which one do you prefer?

kolyshkin: I don't want to overload the error returned with an additional "not sure"; I prefer it the way it is.

manugupt1:
Hi @kolyshkin, would you be able to have a look at it? I would like to get it reviewed and address comments if you have any. I have added things that I have found out; but a final review would be great.

mountinfo/mounted_linux_test.go
```diff
@@ -314,6 +322,38 @@ func TestMountedBy(t *testing.T) {
...
+			if err == nil {
+				if !openat2Supported {
+					// stat does not detect bind-mounts correctly.
+					if tc.isBind && mounted == true {
+						t.Errorf("MountFast: expected mounted as false for bind-mounts, got %v", mounted)
+					}
+					if !tc.isBind && mounted != exp {
+						t.Errorf("MountFast: expected mounted as %v, got %v", exp, mounted)
+					}
+					if sure != tc.isSure {
```
mkimuram, line 336: I understand that this code is tests for `isSure == false` case in openat2 unsupported environment.
According to comment in `mountedFast`, `isSure == false` can be returned for `isMount == false` case or `isBind == true` case.
Therefore, expected `isSure` will be like below:

| openat2Supported  | isMount |  isBind |  expected isSure |    Note                          |
|------------------ |---------| --------| -----------------| ---------------------------------|
|  true             |  true   |   true  |    true          |                                  |
|  true             |  true   |   false |    true          |                                  |
|  true             |  false  |   true  |    true          |  This won't actually happen.     |
|  true             |  false  |   false |    true          |                                  |
|  false            |  true   |   true  |    false         |                                  |
|  false            |  true   |   false |    true          |                                  |
|  false            |  false  |   true  |    false         |  This won't actually happen.     |
|  false            |  false  |  false  |   false          |                                  |

Then, isn't it less confusing not defining `isSure` in each test case, instead making expected `isSure` by using `isMount` and `isBind`? I guess that it can be `isMount && !isBind` when `openat2Supported == false`. Manually defining `isSure` would make it difficult for other people to add additional tests.

manugupt1: Hmm! I think adding `isSure` helps a bit and the test-writer should be able to know if the value they are adding is guaranteed to be a mount-point or not.

There are a couple of cases where `sure` is `false` for `openat2Supported` as well.
For eg: `non-existent path` and `broken_symlink`. Both of them lie in the path of `normalizePath` where is sure is returned as false which is the 3rd row.

I prefer keeping it as it is rather than increasing the branching as this test is really quite complicated.

mkimuram: Oh, I missed to add the error return cases in the table. Thank you for pointing it out.
However, aren't they handled in else clause for `if err == nil`?

Actually, `sure` seems directly compared with `true` and `false` in else clauses just below, it would be confusing to keep `isSure` in the current way. So, if we decide to keep `isSure`, I prefer separately defining something like `isSureNoOpenat2` and `expectMountNoOpenat2` and comparing them with the actual returned values for `openat2Supported == false` case (We may also need to rename `isSure` to `isSureOpenat2` to avoid confusion and compare it for `openat2Supported == true` case).

BTW, `isSure: true` for "non-existent path" case seems incorrect, then.

manugupt1: Yes, it makes sense to separate out `sure tests` for better readability. Yes, they are added there and that's how I realized these cases. I will try to do that.

For the case of `non-existent path`; The stat call leads to the failure: https://github.com/moby/sys/blob/26b84be33cb40ce258b8c616d0a0aa8c250ae356/mountinfo/mounted_unix.go#L41-L44 in normalizePath and hence we are not sure. Its the same as what k8s does today: Ref: https://github.com/kubernetes/kubernetes/blob/34d1d2a4b98e496145d49aee322a94d8387bfdf7/staging/src/k8s.io/mount-utils/mount_linux.go#L337-L340

FWIW, I think the user should not use the `MountedFast` if they want to be absolutely be sure.  They should choose `Mounted`.

jsafrane:
The code looks good to me.

On personal note, having unit tests run as root and depend on kernel version is IMHO bad, as different people (and CI) will test different things. It seems moby is fine with that.

mountinfo/mounted_linux_test.go
mkimuram: Thank you for your fix. It improved a lot.

I think that we can improve a bit by moving `openat2Supported` check to the top level like below (I'm not sure that it's correct, so please double-check):

```golang
// for any error, regardless of kernel version, sure is false and mounted is false.
if err != nil {
	if sure == true {}
	if mounted != false {}
}

if openat2Supported {
	if err == nil {
		if sure != true {}
		if mounted != exp {}
	}
	
	if tc.isMount && sure == false {} 
} else {
	if err == nil {
		// stat does not detect bind-mounts correctly.
		if tc.isBind && mounted == true {}
		if !tc.isBind && mounted != exp {}
	} 

	if tc.isMount {
		if tc.isBind && sure {} 
		if !tc.isBind && !sure {}
	} else {
		// if it is not a mount-point, we are never sure on older kernels.
		if sure {}
	}
}
```

By above refactoring, we can separately focus on the behaviors of openat2Supported case and non-openat2Supported case (Actually, I didn't noticed that we can simplify like it, until I see your refactored code. Very nice work).

Also, `tc.isSure` seems no longer referenced, can we remove `isSure` entirely?

mkimuram: Thank you for handling this.

Should we also make error if `tc.isMount && !sure` and `!tc.isMount && sure` for `err == nil && !openat2Supported && !tc.isBind` case, or line 345?
(Actually, equivalent codes existed in the previous code).

These conditions are exactly what consumers like k8s will rely on whether to use fall back logic, it would be worth checking here. Other than this, it looks good to me.

manugupt1: Sorry! I missed it; my bad; thank you for reviewing so patiently. I need to get better at keeping all the branchings; if you have suggestions I can take it as well. 

I will update it.

mkimuram: Thank you. It looks good to me now.

mountinfo/mounted_linux_test.go
```diff
@@ -324,34 +317,43 @@ func TestMountedBy(t *testing.T) {
...
 			mounted, sure, err := MountedFast(m)
-			if err == nil {
-				if !openat2Supported {
-					// stat does not detect bind-mounts correctly.
-					if tc.isBind && mounted == true {
-						t.Errorf("MountFast: expected mounted as false for bind-mounts, got %v", mounted)
-					}
-					if !tc.isBind && mounted != exp {
+
+			// for any error, regardless of kernel version, sure and mounted is false.
```
kolyshkin, line 321: Nits:
* `s/is/are/`
* `s/for/For/`

mountinfo/mounted_linux_test.go
```diff
@@ -324,34 +317,43 @@ func TestMountedBy(t *testing.T) {
...
+			if err != nil {
+				if sure == true {
+					t.Errorf("MountFast: expected sure to be false, got %v", sure)
+				}
+				if mounted != false {
```
kolyshkin, line 326: The code would be easier to read if you use the same comparison (either `== false` or `!= true`) in both cases.

mountinfo/mounted_linux_test.go
```diff
@@ -324,34 +317,43 @@ func TestMountedBy(t *testing.T) {
...
-					if sure != true {
-						t.Errorf("MountFast: expected sure as true, got %v", sure)
-					}
-					if mounted != exp {
-						t.Errorf("MountFast: expected mounted as %v, got %v", exp, mounted)
+					if tc.isBind {
+						if sure == true {
+							t.Errorf("MountFast: expected sure to be false, got %v", sure)
+						}
+						if mounted != false {
```
kolyshkin, line 342: ditto

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,49 @@ import (
...
+
+	path, err = normalizePath(path)
+	if err != nil {
+		return false, false, err
+	}
+
+	return mountedFast(path)
+}
+
+// mountedFast combines mountedByOpenAt2 and mountedByStat
```
kolyshkin, line 29: nit: missing period.

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,49 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// MountedFast is a method of detecting mount points without reading
+// mountinfo from procfs. A mount-point check is guaranteed to be a
+// mount-point or not only when sure is true. When sure is false, the
```
kolyshkin, line 12: nit: `s/mount-point/mount point/`. Here "mount point" is a common term, see e.g. man 8 mount.

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,49 @@ import (
...
+// caller needs to check for other methods (eg: parse /proc/mounts)
+// to successfully determine if it is a mount-point.
```
kolyshkin, line 14: ditto

kolyshkin: Also, probably makes sense to say this function is only available on Linux.

kolyshkin:
I think we're almost good to go.

Please
 * merge the last two commits
 * change "mount-point" to "mount point" everywhere.

manugupt1:
@kolyshkin Would you be able to have a final go and check it please?

jingxu97:
@kolyshkin could you help approve this? Thanks!

kolyshkin:
I started to re-review this, and ended up fixing various nits myself. Please see #100 🙏🏻

thaJeztah:
This was carried in https://github.com/moby/sys/pull/100 and now merged, so closing this one

thanks for contributing and reviewing, @manugupt1 !

PR #100, opened 2022-01-14
mountinfo: add MountedFast

kolyshkin:
This is a carry of #97 -- mostly the same code, arranged slightly differently. I have also cleaned up tests a bit.

@manugupt1 PTAL

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -7,6 +7,34 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// MountedFast is a method of detecting a mount point without reading
+// mountinfo from procfs. A caller can only trust the result if no error
+// and sure == true are returned. Otherwise, other methods (e.g. parsing
+// /proc/mounts) have to be used. If unsure, use Mounted instead (which
+// uses MountedFast, but falls back to parsing mountinfo if needed).
+//
+// If a non-existent path is specified, an appropriate error is returned.
+// In case the caller is not interested in this particular error, it should
+// be handled separately using e.g. errors.Is(err, os.ErrNotExist).
+//
+// This function is only available on Linux. When available (since kernel
+// v5.6), openat2(2) syscall is used to reliably detect all mounts. Otherwise,
+// the implementation falls back to using stat(2), which can reliably detect
+// normal (but not bind) mounts.
+func MountedFast(path string) (mounted, sure bool, err error) {
+	// Root is always mounted.
+	if path == string(os.PathSeparator) {
+		return true, true, nil
+	}
+
+	path, err = normalizePath(path)
+	if err != nil {
+		return false, false, err
+	}
+	mounted, sure, err = mountedFast(path)
+	return
+}
+
 // mountedByOpenat2 is a method of detecting a mount that works for all kinds
 // of mounts (incl. bind mounts), but requires a recent (v5.6+) linux kernel.
 func mountedByOpenat2(path string) (bool, error) {
@@ -34,24 +62,40 @@ func mountedByOpenat2(path string) (bool, error) {
 	return false, &os.PathError{Op: "openat2", Path: path, Err: err}
 }
 
-func mounted(path string) (bool, error) {
-	path, err := normalizePath(path)
-	if err != nil {
-		return false, err
+// mountedFast is similar to MountedFast, except it expects a normalized path.
+func mountedFast(path string) (mounted, sure bool, err error) {
+	// Root is always mounted.
+	if path == string(os.PathSeparator) {
+		return true, true, nil
 	}
+
 	// Try a fast path, using openat2() with RESOLVE_NO_XDEV.
-	mounted, err := mountedByOpenat2(path)
+	mounted, err = mountedByOpenat2(path)
 	if err == nil {
-		return mounted, nil
+		return mounted, true, nil
 	}
+
 	// Another fast path: compare st.st_dev fields.
 	mounted, err = mountedByStat(path)
 	// This does not work for bind mounts, so false negative
 	// is possible, therefore only trust if return is true.
 	if mounted && err == nil {
+		return true, true, nil
+	}
+
+	return
+}
+
+func mounted(path string) (bool, error) {
+	path, err := normalizePath(path)
+	if err != nil {
+		return false, err
+	}
+	mounted, sure, err := mountedFast(path)
+	if sure && err == nil {
 		return mounted, nil
 	}
 
-	// Fallback to parsing mountinfo
+	// Fallback to parsing mountinfo.
 	return mountedByMountinfo(path)
 }
--- a/mountinfo/mounted_linux_test.go
+++ b/mountinfo/mounted_linux_test.go
@@ -34,7 +34,7 @@ func tMount(t *testing.T, src, dst, fstype string, flags uintptr, options string
 	return nil
 }
 
-var testMounts = []struct {
+type testMount struct {
 	desc       string
 	isNotExist bool
 	isMount    bool
@@ -48,7 +48,9 @@ var testMounts = []struct {
 	// simplicity (no need to check for errors or call t.Cleanup()), and
 	// it may call t.Fatal, but practically we don't expect it.
 	prepare func(t *testing.T) (string, error)
-}{
+}
+
+var testMounts = []testMount{
 	{
 		desc:       "non-existent path",
 		isNotExist: true,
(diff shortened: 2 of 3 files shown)


mountinfo/mounted_linux.go
```diff
@@ -34,24 +49,50 @@ func mountedByOpenat2(path string) (bool, error) {
...
+	}
+
+	return
+}
+
+func mounted(path string) (bool, error) {
+	path, mounted, sure, err := mountedFast(path)
+	if sure {
+		return mounted, err
 	}
```
manugupt1, line 94: I slightly prefer having this check as (even when we have tests) as an extra check.

```
if sure && err == nil 
```

manugupt1:
One nit and this looks awesome! thanks!

mountinfo/mounted_linux.go
kolyshkin: Sure.

Also, I just realized that if `normalizePath` returns an error, it does not make sense to continue with parsing mountinfo, so this code needs to be changed anyway.

Will fix.

kolyshkin:
Should we maybe rename this to `FastMounted()`? @thaJeztah WDYT?

manugupt1:
> Should we maybe rename this to `FastMounted()`? @thaJeztah WDYT?

I like `MountedFast` better because it reads as `MountedByFastMethods` to me; just like `MountedByOpenAt2` and `MountedByStat` in the rest of the code base.

mountinfo/mounted_linux_test.go
```diff
@@ -385,3 +421,22 @@ func TestMountedByOpenat2VsMountinfo(t *testing.T) {
 		}
 	}
 }
+
+func TestRootIsMounted(t *testing.T) {
```
manugupt1, line 425: Is there a reason that we need a separate test here? 
Can we not  make it part of `testMounts` in `TestMountedBy`?
This will help increase coverage for both stat and openat2 as well.

kolyshkin: Because root mount ("/") is a very separate case, it is handled by Mounted / MountedFast separately, and neither `mountedByStat` nor `mountedByOpenat2` can work with it.

manugupt1 (approved):

kolyshkin:
@thaJeztah @cpuguy83 PTAL

mountinfo/mounted_linux.go
```diff
@@ -34,24 +63,41 @@ func mountedByOpenat2(path string) (bool, error) {
...
+		return true, true, nil
 	}
+
 	// Try a fast path, using openat2() with RESOLVE_NO_XDEV.
-	mounted, err := mountedByOpenat2(path)
+	mounted, err = mountedByOpenat2(path)
 	if err == nil {
-		return mounted, nil
+		sure = true
+		return
```
thaJeztah, line 77: I'm not a fan of "naked" returns; would you be ok with changing this to explicitly return the variables?

```
return mounted, sure, nil
```

or

```
return mounted, true, nil
```

mountinfo/mounted_linux.go
```diff
@@ -34,24 +63,41 @@ func mountedByOpenat2(path string) (bool, error) {
...
+
 	// Another fast path: compare st.st_dev fields.
 	mounted, err = mountedByStat(path)
 	// This does not work for bind mounts, so false negative
 	// is possible, therefore only trust if return is true.
 	if mounted && err == nil {
+		sure = true
+	}
+
+	return
```
thaJeztah, line 88: same here

mountinfo/mounted_linux_test.go
```diff
@@ -314,6 +316,47 @@ func TestMountedBy(t *testing.T) {
...
+				}
+				if sure {
+					t.Error("MountedFast: expected sure to be false on error")
+				}
+				if mounted {
+					t.Error("MountedFast: expected mounted to be false on error")
+				}
+			} else {
+				if openat2Supported {
+					if mounted != exp {
```
thaJeztah, line 335: This test is becoming quite complex; could we remove the local `exp` variable, and instead use `tc.isMount` ? (I had to scroll up to learn what `exp` was)

mountinfo/mounted_linux_test.go
```diff
@@ -314,6 +316,47 @@ func TestMountedBy(t *testing.T) {
 				}
 			}
 
+			// Check the public MountedFast() function as a whole.
+			mounted, sure, err := MountedFast(m)
```
thaJeztah, line 320: Also see my other comment about this test becoming quite complex; wondering if we should just move this to a separate (`TestMountedFast`) test. There will be _some_ boilerplating / code duplication, but perhaps it's a good trade-off compared to the complexity we're ending up into now. i.e. using `subtests` normally avoids having to include the function name in the `t.Errorf`, but because we're now combining so many we have to do it again.

mountinfo/mounted_linux_test.go
thaJeztah: Perhaps worth adding a short GoDoc to describe that, e.g. (suggestion welcome; it's just a quick blurb)

```go
// TestRootIsMounted verifies the special handling for mounting root ("/")
```

I'd be in favor of naming the test `TestMountedXXXX`; it helps filtering tests

thaJeztah:
> > Should we maybe rename this to `FastMounted()`? @thaJeztah WDYT?
> 
> I like `MountedFast` better because it reads as `MountedByFastMethods` to me; just like `MountedByOpenAt2` and `MountedByStat` in the rest of the code base.

I agree; let's stick with `MountedXXX`

thaJeztah (changes requested):
left some comments/suggestions

mountinfo/mounted_linux_test.go
kolyshkin: done

mountinfo/mounted_linux.go
kolyshkin: Makes sense, done.

mountinfo/mounted_linux.go
kolyshkin: done

mountinfo/mounted_linux_test.go
kolyshkin: Yes, this looks better. Updated.

mountinfo/mounted_linux_test.go
kolyshkin: fixed

kolyshkin:
@manugupt1 @thaJeztah PTAL 🙏🏻

mountinfo/mounted_linux.go
```diff
@@ -7,6 +7,35 @@ import (
...
+// /proc/mounts) have to be used. If unsure, use Mounted instead (which
+// uses MountedFast, but falls back to parsing mountinfo if needed).
+//
+// If a non-existent path is specified, an appropriate error (rather than
+// "not mounted") is returned. In case the caller is not interested in this
+// particular error, it should be handled separately using e.g.
+// errors.Is(err, os.ErrNotExist).
+//
+// This function is only available on Linux. The implementation mostly
+// relies on openat2(2) syscall, available since Linux kernel v5.6.
```
manugupt1, line 22: nit: this reads to me that we rely on openat2. In reality, we chose the sys call based on the linux version. Would you be able to update the language bit. 

The implementation relies on openat2(2) sys call when available. When it is not, it will use stat(2) syscall which can detect normal (but not bind) mounts.

manugupt1 (changes requested):

thaJeztah:
looks good to me after @manugupt1's suggestion above has been looked at

mountinfo/mounted_linux.go
kolyshkin: I'm not sure how to spell it out so the kernel version information is not lost -- and I think it makes sense to mention it.

Ended up with

```go
// This function is only available on Linux. When available (since kernel
// v5.6), openat2(2) syscall is used to reliably detect all mounts. Otherwise,
// the implementation falls back to using stat(2), which can reliably detect
// normal (but not bind) mounts.
```

manugupt1: This sounds much better to me. Thanks

kolyshkin:
@thaJeztah @moby/moby-maintainers PTAL

kolyshkin:
One thing I failed to document is shadowed mounts.

For example, if `/mnt/one` is a mount point, but later sysadmin does `mount /dev/sda /mnt`, the `/mnt/one` path becomes shadowed by the new mount. The shadowed mount is still visible from mountinfo, but it is impossible to access it (for example, to unmount).

`MounedFast` returns `ENOENT` for such shadowed mounts, since the path is inaccessible. `Mounted` used to return "true" for such mounts, but since commit dbd468bb51eb9db344 it also returns `ENOENT`.

I am failing to describe all this in a nice yet compact way.

In any case, this can be done later.

manugupt1 (approved):
thank you for doing this in such a nice way and helping me. I learnt a lot from you.

thaJeztah (approved):
LGTM


we can further improve documentation in follow ups if we see a need 👍

manugupt1:
@kolyshkin @thaJeztah 
Would you be able to cut a release for Mount I for with this change?

Thanks

thaJeztah:
Yes, I wanted to prose doing so.

Needed to check again in which order things had to be released (as the mountinfo and mount packages have a dependency on each other)

kolyshkin:
https://github.com/moby/sys/releases/tag/mountinfo%2Fv0.6.0

kolyshkin:
> (as the mountinfo and mount packages have a dependency on each other)

Only `mount` depends on `mountinfo`, but not vice versa.

AFAIK we don't need to do anything; updating the mount's go.mod to require mountinfo/v.0.6.0 is highly optional (but makes sense to do before its next release, of course).

thaJeztah:
Yes, it's good to update them, so that projects that use them are using the same version

Issue #99, opened 2022-01-14
CI: mac os x linter warnings ignored


kolyshkin:
I have (accudentally) noticed that on Mac OS X the warnings from golangci-lint are ignored. For example, see 
https://github.com/kolyshkin/sys/runs/4823094862?check_suite_focus=true

This can be caused by shell differences and the use of subshell by Makefile here:

https://github.com/moby/sys/blob/03b9f8d59a07f5206a2264105f4903a222aea964/Makefile#L19

Alas I don't have a Mac to debug this. @thaJeztah maybe you could take a look?

kolyshkin:
Sorry, I meant this place: 
https://github.com/moby/sys/blob/03b9f8d59a07f5206a2264105f4903a222aea964/Makefile#L37-L38

I am 95% sure golangci-lint exits with a non-zero code here, and yet it's not returned by make.

thaJeztah:
Not at my computer right now, but reminds me of https://github.com/moby/sys/pull/19, which was meant to fix something similar?

thaJeztah:
So it looks to fail on my machine (at commit 03b9f8d59a07f5206a2264105f4903a222aea964), so not sure why it wouldn't fail in CI 🤔 

```bash
make lint
_build/bin/golangci-lint version
golangci-lint has version 1.43.0 built from 861262b7 on 2021-11-03T11:57:46Z
for p in mountinfo mount signal symlink; do \
		(cd $p && go mod download \
		&& ../_build/bin/golangci-lint run); \
	done
ERRO Running error: context loading failed: failed to load packages: failed to load with go/packages: err: exit status 1: stderr: go: inconsistent vendoring in /Users/sebastiaan/go/src/github.com/moby/sys/mountinfo:
	golang.org/x/sys@v0.0.0-20211025201205-69cdffdb9359: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt

	To ignore the vendor directory, use -mod=readonly or -mod=mod.
	To sync the vendor directory, run:
		go mod vendor
ERRO Running error: context loading failed: failed to load packages: failed to load with go/packages: err: exit status 1: stderr: go: inconsistent vendoring in /Users/sebastiaan/go/src/github.com/moby/sys/mount:
	github.com/moby/sys/mountinfo@v0.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
[...]

PR #101, opened 2022-01-14
mount: fix unused/deadcode warnings on Mac

kolyshkin:
Since commit 7a521626bc899dc Mount is not implemented on Darwin, so
mount() is left unused.

Also, mount errors are only used from mount(), and thus are not used on
Darwin.

This fixes a bunch of warnings from "unused" and "deadcode" lints.

Alas, those warnings did not result in CI failures on Mac; see https://github.com/moby/sys/issues/99

### DIFF
--- a/mount/mount_errors.go
+++ b/mount/mount_errors.go
@@ -1,5 +1,5 @@
-//go:build !windows
-// +build !windows
+//go:build !darwin && !windows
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
LGTM, thanks!

PR #105, opened 2022-02-05
Makefile: rm .SHELLFLAGS, add set -e

kolyshkin:
Apparently GHA CI Mac OS X environment uses old GNU Make (v3.81) which
does not yet support `.SHELLFLAGS` (added by commit 8589e2e7).

Since we call a few commands (under `for`) from the make target,
intermediate failures are ignored unless `-e` shell option is set (or
explicit `&&` is used, which is not possible to do in `for`).

Example:

```sh
for cmd in false true; do $cmd; done
```
will succeed unless `set -e` is set, because the last command executed is `true` and its exit code is `0`.

To fix, use explicit `set -e` where we execute more than a single command.

Also, replace `&&` with `;` for multiple commands. Technically this is not
required for the fix, but since having `set -e` results in implicit `&&` everywhere,
there is no need to have `&&` explicitly. In other words, this change is
more like a reminder that we do not have to use `&&` here.

Fixes: 8589e2e7

Fixes: #99

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-.SHELLFLAGS = -ec
 PACKAGES ?= mountinfo mount signal symlink
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
@@ -15,8 +14,9 @@ clean:
 
 .PHONY: test
 test: test-local
+	set -eu; \
 	for p in $(PACKAGES); do \
-		(cd $$p && go test $(RUN_VIA_SUDO) -v .); \
+		(cd $$p; go test $(RUN_VIA_SUDO) -v .); \
 	done
 
 # Test the mount module against the local mountinfo source code instead of the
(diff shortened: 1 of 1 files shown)


kolyshkin:
More thoughts on topic:

1. We have to use `cd` from Makefile, because `go` tools (and `golangci-lint`) treat current directory in a special way, and does not support ways to specify an alternative directory.

2. Traditionally, such issues are solved with sub-Makefiles, but I don't like it for this repo (since it's easy, and maintaining more Makefiles, even trivial, seems more of a mess than a few `for`s in the single Makefile). I might reconsider that if we'll have more similar issues.

3. (Not really related to this fix, but) there are a few alternatives to using subshell to cd back to the top-level directory. They all have their deficiencies, for example:
 - `pushd`/`popd` won't work with dash (the default sh on Ubuntu);
 - using `cd $pkgdir; test; cd ..` would not work in case the hierarchy of packages is not flat (e.g. `PACKAGES = mount mountinfo symlink nice/one nice/two`); currently this is not the case but being future-proof feels nice;
 - using `$PWD` (or `pwd`) in some way (e.g. `TOPDIR=$$PWD; ...; cd $TOPDIR`) is probably OK, too, but maybe it looks way too verbose.
 
Using subshell, of course, is not free from deficiencies, too:
 - it's not the best way in terms of performance (but we don't care in this case);
 - the "cd .." is implicit, which I guess hurts readability;
 - it might look cryptic to people not very familiar with shell syntax.

kolyshkin:
Fedora job failed because of dnf repo glitch, and apparently the retry doesn't work :(

CI restarted.

thaJeztah:
Yeah, not sure if I have better options; something like `make -C` / `$(MAKE) -C`, but then we need to do trickery as well to still make it read the Makefile from the "parent" directory. Of course we could define targets for each module, and call those, but that would produce a lot of duplication as well.

I guess this change is ok (at least it doesn't make things _worse_)

thaJeztah (approved):
LGTM

Issue #45, opened 2020-10-02
mount: Add test coverage for RecursiveUnmount


thaJeztah:
While there's tests for functions _used_ by this function (`GetMounts()`, `Unmount()`), there's currently no test coverage for `RecursiveUnmount()`. We could use some test coverage for this function.

Opening this as a tracking issue, and adding the "help wanted" label, in case someone wants to work on this (please comment below if you want to work on this!)

/cc @kolyshkin

kolyshkin:
This is partially addressed by #64.

manugupt1:
@thaJeztah can we close this now?

kolyshkin:
Yes, I think this is now addressed by #64 and #107.

PR #107, opened 2022-03-07
Add test for RecursiveUnmount when a submount fails to unmount.

manugupt1:
Add test for RecursiveUnmount when a submount fails to unmount.

https://github.com/moby/sys/issues/45

Signed-off-by: Manu Gupta <manugupt1@gmail.com>

### DIFF
--- a/mount/mount_unix_test.go
+++ b/mount/mount_unix_test.go
@@ -4,12 +4,14 @@
 package mount
 
 import (
+	"errors"
 	"os"
 	"path"
 	"strings"
 	"testing"
 
 	"github.com/moby/sys/mountinfo"
+	"golang.org/x/sys/unix"
 )
 
 func TestMountOptionsParsing(t *testing.T) {
@@ -228,5 +230,47 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
 	}
 	if !mounted {
 		t.Fatal("expected dir-other to be mounted, but it's not")
+	}
+}
+
+func TestRecursiveUnmount_SubMountFailsToUnmount(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("root required")
+	}
+
+	var (
+		tmp        = t.TempDir()
+		parent     = tmp + "/sub1"
+		child      = tmp + "/sub1/sub2"
+		grandChild = tmp + "/sub1/sub2/sub3"
+	)
+
+	err := os.MkdirAll(grandChild, 0o700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a set of mounts that should result in RecursiveUnmount failure,
+	// caused by the fact that the grandchild mount is shadowed by the child mount,
+	// and the child mount is shadowed by the parent mount. So. these two mounts
+	// are listed in mountinfo, but since they are unreachable, unmount will fail.
+	toMount := []string{grandChild, child, parent}
+	for _, dir := range toMount {
+		dir := dir
+		if err := Mount("tmpfs", dir, "tmpfs", ""); err != nil {
+			t.Fatal(err)
+		}
+		defer Unmount(dir) //nolint:errcheck
+	}
+
+	// unmount shadowed mounts
+	shadowedMounts := []string{child, grandChild}
+	for _, shadowedMount := range shadowedMounts {
+		t.Run(shadowedMount, func(t *testing.T) {
+			err := RecursiveUnmount(shadowedMount)
+			if !errors.Is(err, unix.ENOENT) {
+				t.Fatalf("expected submount(shadowed) %s to return unix.ENOENT, got %v", shadowedMount, err)
+			}
+		})
 	}
 }


mount/mount_unix.go
```diff
@@ -21,7 +21,7 @@ func Mount(device, target, mType, options string) error {
 
 // Unmount lazily unmounts a filesystem on supported platforms, otherwise does
 // a normal unmount. If target is not a mount point, no error is returned.
-func Unmount(target string) error {
+var Unmount = func(target string) error {
```
kolyshkin, line 24: I'd rather not do that for the sake of a single unit test.

Unfortunately I don't know of any other way to make the deferred unmount fail.

kolyshkin: Hmm, maybe I do.

Try to create a mount which is shadowed by another mount. This will be shown in mountinfo, but the directory will not be there, so the unmount will fail with ENOENT.

IOW something like
1. mount "tmp/dir1/subdir" using tmpfs.
2. mount "tmp/dir1" using tmpfs.

Now, RecursiveUnmount will try to unmount "tmp/dir1/subdir", and since it's hidden by the second mount, its unmount will fail with ENOENT.

mount/mount_unix_test.go
```diff
@@ -251,3 +252,42 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+		if err := os.MkdirAll(dir, 0o700); err != nil {
+			t.Fatal(err)
+		}
+		if err := Mount("tmpfs", dir, "tmpfs", ""); err != nil {
+			t.Fatal(err)
+		}
+		//nolint:errcheck
+		defer Unmount(dir)
+	}
+
```
kolyshkin, line 281: I see that you copied the "TooGreedy" test except for its sanity check part. Do you think it's not necessary?

mount/mount_unix.go
manugupt1: Thanks for the suggestion. I did not know about shadow mounts. I updated the test by using your suggestion.

manugupt1:
HI @kolyshkin 
Will you be able to review this when you get a chance?

Thanks

kolyshkin:
LGTM, except I don't understand why 3 mounts are needed -- it seems that 2 are enough.

One other thing, can you please add a short description of what the added case tests (something like "$NAME checks that RecursiveUnmount does return an error when a submount fails to be unmounted")

manugupt1:
> LGTM, except I don't understand why 3 mounts are needed -- it seems that 2 are enough.
> 
> One other thing, can you please add a short description of what the added case tests (something like "$NAME checks that RecursiveUnmount does return an error when a submount fails to be unmounted")

Having 3 mounts ensures that we check for the cases when suberr is nil and not nil.

mount/mount_unix_test.go
```diff
@@ -230,3 +234,43 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+	// unmount shadowed mounts
+	shadowedMounts := []string{child, grandchild}
+	for _, shadowedMount := range shadowedMounts {
+		// Unmount dir, make sure dir-other is still mounted.
```
kolyshkin, line 269: This comment no longer reflects reality -- I don't see "dir" and "dir-other" in here.

mount/mount_unix_test.go
```diff
@@ -230,3 +234,43 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+	err := os.MkdirAll(path.Join(tmp, dir), 0o700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	grandchild := path.Join(tmp, dir)
+	child := filepath.Dir(grandchild)
+	parent := filepath.Dir(child)
+
+	// order is necessary to create shadow mounts.
```
kolyshkin, line 255: I guess it makes sense to describe the idea here, so that the future readers will get a grasp about what is going on.

```go 
// Create a set of mounts that should result in RecursiveUnmount failure,
// caused by the fact that the grandchild mount is shadowed by the child mount,
// and the child mount is shadowed by the parent mount. So. these two mounts
// are listed in mountinfo, but since they are unreachable, unmount will fail.
```

mount/mount_unix_test.go
```diff
@@ -4,11 +4,15 @@
...
+	"errors"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/moby/sys/mountinfo"
```
thaJeztah, line 16: Minor nit; can you remove the empty line and sort the imports?

thaJeztah:
@manugupt1 with https://github.com/moby/sys/pull/109 merged, do you need new releases of `mount` and `mountinfo` after this was merged, or do you anticipate other changes coming before we should do a new release?

kolyshkin:
> I do not know enough k8s to see if I can base on the latest commit over master and then do follow ups either here or there if reqd.

Technically you can, but maintainers will complain. For some package which did not do a release for a long time they can make an exception, and of course you can use an untagged version for testing.

So, let us know once you finished the testing and need a release, and we'll tag one.

mount/mount_unix_test.go
```diff
@@ -230,3 +233,47 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		//nolint:errcheck
+		defer Unmount(dir)
+	}
+
+	// unmount shadowed mounts
+	shadowedMounts := []string{child, grandchild}
+	for _, shadowedMount := range shadowedMounts {
+		t.Run("RecursiveUnmount returns an error", func(t *testing.T) {
```
kolyshkin, line 271: So, with this t.Run we have two subtests which are similarly named. I guess either to remove t.Run, or make the subtest name different (e.g. by `t.Run(shadowedMount)`).

@thaJeztah WDYT?

mount/mount_unix_test.go
```diff
@@ -230,3 +233,47 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+	// and the child mount is shadowed by the parent mount. So. these two mounts
+	// are listed in mountinfo, but since they are unreachable, unmount will fail.
+	toMount := []string{grandchild, child, parent}
+	for _, dir := range toMount {
+		dir := dir
+		if err := Mount("tmpfs", dir, "tmpfs", ""); err != nil {
+			t.Fatal(err)
+		}
+		//nolint:errcheck
+		defer Unmount(dir)
```
thaJeztah, line 265: Didn't leave it as a comment earlier as it's a minor nit, but if you're updating, perhaps put the `//nolint` after the line; slightly cleaner, and makes it slightly clearer which line the `//nolint` applies to.

```go
defer Unmount(dir) //nolint:errcheck
```

mount/mount_unix_test.go
thaJeztah: Good catch; yes, I think that would work; I'd change this to:

```go
		t.Run(shadowedMount, func(t *testing.T) {
```

mount/mount_unix_test.go
```diff
@@ -230,3 +233,47 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+
+	// unmount shadowed mounts
+	shadowedMounts := []string{child, grandchild}
+	for _, shadowedMount := range shadowedMounts {
+		t.Run("RecursiveUnmount returns an error", func(t *testing.T) {
+			if err := RecursiveUnmount(shadowedMount); err == nil {
+				t.Fatalf("expected submount %s to return an error", shadowedMount)
+			} else if err != nil && !errors.Is(err, unix.ENOENT) {
+				t.Fatalf("expected submount(shadowed) %s to return unix.ENOENT", shadowedMount)
+			}
```
thaJeztah, line 276: Given that `errors.Is()` already returns `false` if `err == nil`, this can be simplified to:

```suggestion
			err := RecursiveUnmount(shadowedMount)
			if !errors.Is(err, unix.ENOENT) {
				t.Fatalf("expected submount(shadowed) %s to return unix.ENOENT, got %v", shadowedMount, err)
			}
```

mount/mount_unix_test.go
```diff
@@ -230,3 +233,47 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+
+	dir := "sub1/sub2/sub3"
+	err := os.MkdirAll(path.Join(tmp, dir), 0o700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	grandchild := path.Join(tmp, dir)
+	child := filepath.Dir(grandchild)
+	parent := filepath.Dir(child)
```
thaJeztah, line 252: This seems making things more complicated than needed to find paths we already know what they'll be (relative to `tmp` that is). Given that this only runs on unix/linux, we don't strictly need `filepath` (path separator will always be `/`) and because these paths are "fixed", we don't need to care about path-escaping etc. (also `grandchild` should probably be `grandChild` to be properly camelCase); 

perhaps something like:

```go
var (
	tmp        = t.TempDir()
	parent     = tmp + "/sub1"
	child      = tmp + "/sub1/sub2"
	grandChild = tmp + "/sub1/sub2/sub3"
)

err := os.MkdirAll(grandChild, 0o700)
if err != nil {
	t.Fatal(err)
}
```

That way, it's obvious at a glance what paths we're dealing with (the `var(    )` block makes them align nicely, so it's easier to read for this).

thaJeztah:
Thanks for updating; changes look good - could you squash the commits? (it's fine to squash my suggestions with your commit 👍)

thaJeztah (approved):
LGTM, thanks!

@kolyshkin PTAL

kolyshkin (approved):
LGTM

PR #112, opened 2022-04-17
Correctly identify shadow mounts for Mounted Function

manugupt1:
fix: Correctly identify shadow mounts for Mounted Function.
normalizePath can return ENOENT when it is not a mountpoint.
When ENOENT is returned, assume that a shadow mount can
exist and fallback on reading /proc/mounts.

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -1,6 +1,8 @@
 package mountinfo
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 	"path/filepath"
 
@@ -87,15 +89,18 @@ func mountedFast(path string) (mounted, sure bool, err error) {
 }
 
 func mounted(path string) (bool, error) {
-	path, err := normalizePath(path)
+	normalizedPath, err := normalizePath(path)
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
+		return mountedByMountinfo(path)
+	}
 	if err != nil {
 		return false, err
 	}
-	mounted, sure, err := mountedFast(path)
+	mounted, sure, err := mountedFast(normalizedPath)
 	if sure && err == nil {
 		return mounted, nil
 	}
 
 	// Fallback to parsing mountinfo.
-	return mountedByMountinfo(path)
+	return mountedByMountinfo(normalizedPath)
 }
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -91,7 +91,15 @@ func mountedFast(path string) (mounted, sure bool, err error) {
 func mounted(path string) (bool, error) {
 	normalizedPath, err := normalizePath(path)
 	if err != nil && errors.Is(err, fs.ErrNotExist) {
-		return mountedByMountinfo(path)
+		if filepath.IsAbs(path) {
+			return mountedByMountinfo(path)
+		} else {
+			if absPath, err := filepath.Abs(path); err != nil {
+				return false, err
+			} else {
+				return mountedByMountinfo(absPath)
+			}
+		}
 	}
 	if err != nil {
 		return false, err
--- a/mountinfo/mounted_linux_test.go
+++ b/mountinfo/mounted_linux_test.go
@@ -455,3 +455,36 @@ func TestMountedRoot(t *testing.T) {
 		}
 	}
 }
+
+func TestMountedShadowMount(t *testing.T) {
+	if os.Getuid() != 0 {
+		t.Skip("root required")
+	}
+
+	var (
+		tmp                = t.TempDir()
+		mount              = tmp + "/sub1"
+		shadowedMountpoint = tmp + "/sub1/sub2"
+	)
+
+	err := os.MkdirAll(shadowedMountpoint, 0o700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a mountpoint that will be shadowed.
+	if err := tMount(t, "tmpfs", shadowedMountpoint, "tmpfs", 0, ""); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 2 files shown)


manugupt1:
This is not true, because of shadow mounts ( I forgot about that edge-case) but shows a bug in the library. Should I fall back to reading /proc/mounts for shadow mounts.

cc @thaJeztah @kolyshkin

mountinfo/mounted_linux_test.go
```diff
@@ -455,3 +455,40 @@ func TestMountedRoot(t *testing.T) {
...
+		mount       = tmp + "/sub1"
+		shadowMount = tmp + "/sub1/sub2"
+	)
+
+	err := os.MkdirAll(shadowMount, 0o700)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a set of mounts that should result in RecursiveUnmount failure,
```
manugupt1, line 475: Fix comments

mountinfo/mounted_linux.go
```diff
@@ -88,7 +88,7 @@ func mountedFast(path string) (mounted, sure bool, err error) {
 
 func mounted(path string) (bool, error) {
 	path, err := normalizePath(path)
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
```
thaJeztah, line 91: Is `os.IsNotExist()` still the correct way to check this, or should we use `errors.Is(err, os.ErrNotExist)` for this? (not 100% sure, ISTR `os.IsNotExist()` also takes some other cases into account)

mountinfo/mounted_unix.go
```diff
@@ -32,13 +32,13 @@ func mountedByStat(path string) (bool, error) {
...
 	}
 	if realPath, err = filepath.EvalSymlinks(realPath); err != nil {
-		return "", err
+		return path, err
 	}
 	if _, err := os.Stat(realPath); err != nil {
-		return "", err
+		return path, err
 	}
 	return realPath, nil
```
thaJeztah, line 43: Can `filepath.Abs()` or `filepath.EvalSymlinks()` also return an `os.ErrNotExist` error for other reasons? e.g. `filepath.Abs()` calls out to `os.GetCWD()`, which can return an error; https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/os/getwd.go#L22

And I assume `filepath.EvalSymlinks()`  can return an error if any symlink failed to be resolved.

Mostly wondering if `path` is "safe to consume" in all of these error situations. It may be a bit controversial to return _both_ an error _and_ a value. Not sure if there's problematic cases 🤔. 

Not sure what's best in that case though; should the caller use the "original" (non-normalised) path? (concerned if the "normalised" path in that case may contain some path that cannot / should not be used (could be an empty string returned by any of the above?)

mountinfo/mounted_linux.go
manugupt1: It should be errors.Is(err, fs.ErrNotExist) as per documentation.
Ref: https://pkg.go.dev/os#IsNotExist. I modified the code for that.

mountinfo/mounted_unix.go
manugupt1: You are correct, it can fail. I thought about this a bit and made the following code changes.
When `normalizePath` fails or returns an error, we should return `"", err`. Now the only variable is the original path.

I think when ENOENT occurs, we should use the non-normalised path and read from /proc/mounts. This is because /proc/mounts will always have cannonicalized Path. These are the three cases I could think of:
1. path is a absolute path that returned ENOENT: The entry will not exist in /proc/mounts and return false
2. path is a symlink that returned ENOENT: The entry to the symlink will not exist in /proc/mounts and return false.
3. path is not-absolute: If filepath is not absolute, then we can try to get an Absolute filepath and check /proc/mounts. If an error is thrown, we can return it. 

wdyt? I made the changes though.

mountinfo/mounted_linux.go
```diff
@@ -87,15 +89,26 @@ func mountedFast(path string) (mounted, sure bool, err error) {
 }
 
 func mounted(path string) (bool, error) {
-	path, err := normalizePath(path)
+	normalizedPath, err := normalizePath(path)
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
+		if filepath.IsAbs(path) {
```
manugupt1, line 94: This is primarily because we do not know where the ENOENT is coming from; I am not sure I like this as much as I want to :/

kolyshkin:
Is there a specific scenario in which this PR helps?

Currently, `Mount` documentation explicitly states:

> If a non-existent path is specified, an appropriate error is returned.

and this is how it works.

kolyshkin:
I chose to do it that way (returning ENOENT rather than falling back to parsing mountinfo) because parsing mountinfo is super slow operation which can be avoided by all means. OTOH knowing that this non-existent path in fact is a mount won't do any good, because there is nothing we can do with this path -- it can't be unmounted, we can't see the files in there, etc.

This is why I asked if there's a specific issue this PR fixes, so we can assess the severity of such issue versus the added overhead of parsing /proc/mountinfo when a non-existent path is provided.

manugupt1:
> I chose to do it that way (returning ENOENT rather than falling back to parsing mountinfo) because parsing mountinfo is super slow operation which can be avoided by all means. OTOH knowing that this non-existent path in fact is a mount won't do any good, because there is nothing we can do with this path -- it can't be unmounted, we can't see the files in there, etc.
> 
> This is why I asked if there's a specific issue this PR fixes, so we can assess the severity of such issue versus the added overhead of parsing /proc/mountinfo when a non-existent path is provided.

Ah okay! thank you for explaining. I will close this. I did not have a use-case but since I had worked on a similar thing before I thought this would be useful.

PR #115, opened 2022-04-26
Update doc to use fs.ErrNotExist

manugupt1:
There are two nits:
1. Since go 1.18, the doc recommends errors.Is(err, fs.ErrNotExist) rather than errors.Is(err, os.ErrNotExist); hence the doc update.

### DIFF
--- a/mountinfo/mounted_linux.go
+++ b/mountinfo/mounted_linux.go
@@ -15,7 +15,7 @@ import (
 //
 // If a non-existent path is specified, an appropriate error is returned.
 // In case the caller is not interested in this particular error, it should
-// be handled separately using e.g. errors.Is(err, os.ErrNotExist).
+// be handled separately using e.g. errors.Is(err, fs.ErrNotExist).
 //
 // This function is only available on Linux. When available (since kernel
 // v5.6), openat2(2) syscall is used to reliably detect all mounts. Otherwise,
--- a/mountinfo/mountinfo.go
+++ b/mountinfo/mountinfo.go
@@ -15,7 +15,7 @@ func GetMounts(f FilterFunc) ([]*Info, error) {
 //
 // If a non-existent path is specified, an appropriate error is returned.
 // In case the caller is not interested in this particular error, it should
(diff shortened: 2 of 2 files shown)


mountinfo/mounted_unix.go
```diff
@@ -38,7 +38,7 @@ func normalizePath(path string) (realPath string, err error) {
 		return "", err
 	}
 	if _, err := os.Stat(realPath); err != nil {
-		return "", err
+		return "", &os.PathError{Op: "stat", Path: realPath, Err: err}
```
kolyshkin, line 41: This is totally redundant; `os.Stat` already returns `PathError`.

https://go.dev/play/p/aaLjEvYTAfV

mountinfo/mounted_linux.go
```diff
@@ -15,7 +15,7 @@ import (
 //
 // If a non-existent path is specified, an appropriate error is returned.
 // In case the caller is not interested in this particular error, it should
-// be handled separately using e.g. errors.Is(err, os.ErrNotExist).
+// be handled separately using e.g. errors.Is(err, fs.ErrNotExist).
```
kolyshkin, line 18: Is `os.ErrNotExist` deprecated yet? If not, it does not make sense to change this.

kolyshkin (changes requested):

mountinfo/mounted_linux.go
manugupt1: It's not deprecated; but its recommended for new code to use errors.Is(err, fs.ErrNotExist).
Ref: https://pkg.go.dev/os#ErrNotExist

If it does not make sense, I can close this PR.  Thank you for taking the time to review.

kolyshkin: Makes sense, thanks!

kolyshkin (approved):

kolyshkin:
@thaJeztah PTAL

thaJeztah (approved):
LGTM

in practice I often use the "os" variant (its just an alias of "fs" now, but saves adding another import ☺️)

PR #114, opened 2022-04-19
mountinfo: BSDs no longer need cgo nor reflect

Red54:
Alternative PR for https://github.com/moby/sys/pull/113

PR #117, opened 2022-05-19
mountinfo_openbsd: Rewrite int8SliceToString with unsafe

Red54:
https://github.com/moby/sys/pull/114#discussion_r876840831

### DIFF
--- a/mountinfo/mountinfo_openbsd.go
+++ b/mountinfo/mountinfo_openbsd.go
@@ -1,16 +1,19 @@
 package mountinfo
 
-import "golang.org/x/sys/unix"
+import (
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
 
 func int8SliceToString(is []int8) string {
-	var bs []byte
-	for _, i := range is {
-		if i == 0 {
+	for i := range is {
+		if is[i] == 0 {
+			is = is[:i]
 			break
 		}
-		bs = append(bs, byte(i))
 	}
-	return string(bs)
+	return *(*string)(unsafe.Pointer(&is))
 }
 
 func getMountinfo(entry *unix.Statfs_t) *Info {


kolyshkin:
I have submitted https://go-review.googlesource.com/c/sys/+/407195 and hope it will be merged soon. Let's wait a couple of days.

kolyshkin:
The alternative is using something like `BytePtrToString((*byte)(unsafe.Pointer(&is)))`

Red54:
```
package main

import (
        "testing"
        "unsafe"

        "golang.org/x/sys/unix"
)

var is = [90]int8{'a', 'b', 'c'}
var es = "abc"

func int8SliceToString(is []int8) string {
        for i := range is {
                if is[i] == 0 {
                        is = is[:i]
                        break
                }
        }
        return *(*string)(unsafe.Pointer(&is))
}

func TestingInt8SliceToString(t *testing.T) {
        s := int8SliceToString(is[:])
        if s != es {
                t.Fail()
        }
}

func BenchmarkInt8SliceToString(b *testing.B) {
        for i := 0; i < b.N; i++ {
                int8SliceToString(is[:])
        }
}

func TestingBytePtrToString(t *testing.T) {
        s := unix.BytePtrToString((*byte)(unsafe.Pointer(&is)))
        if s != es {
                t.Fail()
        }
}

func BenchmarkBytePtrToString(b *testing.B) {
        for i := 0; i < b.N; i++ {
                unix.BytePtrToString((*byte)(unsafe.Pointer(&is)))
        }
}
```
```
$ go test -bench=. -benchmem  
goos: openbsd
goarch: amd64
pkg: bench
cpu: Intel Core Processor (Skylake, IBRS)
BenchmarkInt8SliceToString-2    458439867                2.585 ns/op           0 B/op          0 allocs/op
BenchmarkBytePtrToString-2      194815534                6.158 ns/op           0 B/op          0 allocs/op
PASS
ok      bench   3.317s
$
```
`BytePtrToString((*byte)(unsafe.Pointer(&is)))` is more than twice as slow.

Red54:
I have submitted https://go-review.googlesource.com/c/go/+/407436 and if it is merged there is no need to split bsd into freebsdlike and openbsd.

kolyshkin:
Guess we should go with what I outlined in https://github.com/moby/sys/pull/114#issuecomment-1146658962 instead

PR #118, opened 2022-06-05
Bump x/sys/unix

akhramov:
A newer version of x/sys/unix updates the OpenBSD bindings
which makes the `int8SliceToString` helper function obsolete.

This change updates the x/sys/unix version

PR #119, opened 2022-06-06
ci: add go mod tidy check

kolyshkin:
This step should fail if a PR author bumped or added some dependencies,
or bumped go version in `go.mod`, but forgot to run `go mod tidy`.

Inspired by https://github.com/moby/sys/pull/118#discussion_r890371395

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,10 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
       uses: actions/checkout@v3
+    - name: go mod tidy
+      run: |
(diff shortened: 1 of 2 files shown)


kolyshkin:
Adding an extra commit to check that CI works as intended (supposed to fail with it)

kolyshkin:
OK, `go mod tidy` step fails as expected with the test commit added (see e.g. https://github.com/moby/sys/runs/6761754513?check_suite_focus=true), removing the test commit.

tianon (approved):

thaJeztah (approved):
LGTM

PR #121, opened 2022-08-27
Integrate github.com/docker/docker/pkg/system Sequential utilities

thaJeztah:
### Integrate github.com/docker/docker/pkg/system Sequential utilities

Integrating the "Sequential" utilities from [github.com/docker/docker/pkg/system](https://github.com/moby/moby/tree/3e8573a85a42d4882068bffab7d83f5d93d417e1/pkg/system),
taken from commit [3e8573a85a42d4882068bffab7d83f5d93d417e1](https://github.com/moby/moby/commit/3e8573a85a42d4882068bffab7d83f5d93d417e1)

Steps taken:

    brew install git-filter-repo
    
    cd ~/Projects
    git clone https://github.com/docker/docker.git moby_sys
    cd moby_sys
    
    # commit taken from
    git rev-parse --verify HEAD
    3e8573a85a42d4882068bffab7d83f5d93d417e1
    
    # remove all code, except for 'pkg/system/filesys_*.go', and rename to /sequential
    git filter-repo \
      --path-glob 'pkg/system/filesys_*.go' \
      --path-rename pkg/system/filesys_unix.go:sequential/sequential_unix.go \
      --path-rename pkg/system/filesys_windows.go:sequential/sequential_windows.go
    
    git checkout -b integrate_moby_sequential
    
    git remote add moby_sys ~/Projects/moby_sys
    git fetch moby_sys
    
    # merge the upstream code
    git merge --allow-unrelated-histories --signoff -S moby_sys/master

### sequential: rename funcs to match their "os" equivalent

Rename the functions so that they can act as a drop-in replacement
for their "os" equivalents.

### DIFF
--- a/sequential/doc.go
+++ b/sequential/doc.go
@@ -1,0 +1,15 @@
+// Package sequential provides a set of functions for managing sequential
+// files on Windows.
+//
+// The origin of these functions are the golang OS and windows packages,
+// slightly modified to only cope with files, not directories due to the
+// specific use case.
+//
+// The alteration is to allow a file on Windows to be opened with
+// FILE_FLAG_SEQUENTIAL_SCAN (particular for docker load), to avoid eating
+// the standby list, particularly when accessing large files such as layer.tar.
+//
+// For non-Windows platforms, the package provides wrappers for the equivalents
+// in the os packages. They are passthrough on Unix platforms, and only relevant
+// on Windows.
+package sequential
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES ?= mountinfo mount signal symlink
+PACKAGES ?= mountinfo mount sequential signal symlink
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
 	freebsd/amd64 openbsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
--- a/sequential/go.mod
+++ b/sequential/go.mod
@@ -1,0 +1,5 @@
+module github.com/moby/sys/sequential
+
+go 1.17
+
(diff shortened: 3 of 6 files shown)


thaJeztah:
@kolyshkin @kzys @AkihiroSuda @dims PTAL

dims (approved):
thanks @thaJeztah for saving the commit info from the other repo!

kzys (approved):

samuelkarp (approved):

sequential/doc.go
```diff
@@ -0,0 +1,15 @@
+// Package sequential provides a set of functions for managing sequential
+// files on Windows.
+//
+// The origin of these functions are the golang OS and windows packages,
+// slightly modified to only cope with files, not directories due to the
+// specific use case.
```
AkihiroSuda, line 6: Not new in this PR, but probably we have to replicate the original copyright header

AkihiroSuda (approved):

sequential/doc.go
thaJeztah: Ah, good point; perhaps extract the bits that were copied to a separate file (I'd have to check how much was copied; I think that's only in the `CreateTemp`, which uses similar /same logic as `os.CreateTemp`)

We can look at that in a follow up in deed 👍

thaJeztah: FWIW; was also discussing with @crazy-max to see to what extend we still need that function; it may be that some parts were needed for "pre go1.9" code, which didn't work well with Windows, but some of those reasons may no longer be there, so possibly we can start to remove some of this.

I think it's still useful to first do the migration, and go from there though (proposing to tag it as `v0.5`, so that any deprecation that would happen could be done before `v1.0.0`).

thaJeztah:
I'll bring this one in, and update the PR's in other projects that were opened to move to this module.

If those "look good", I'll tag a `v0.5.0` release

PR #124, opened 2022-09-02
Remove Fedora since Ubuntu's kernel is new enough

kzys:
The Fedora on Vagrant setup was introduced to test syscalls such as openat2. However, GitHub's Ubuntu runner is now having Linux 5.15 even in ubuntu-20.04.

----
(The below is the original PR description)

We've been having Fedora on macOS to test newer kernels. However now Ubuntu is having 5.15, which is 0.1 newer than Fedora 35.

* Ubuntu 20.04 uses`Linux fv-az137-41 5.15.0-1017-azure`
* Ubuntu 22.04 uses `Linux fv-az267-637 5.15.0-1017-azure`
* Fedora 35 uses `Linux fedora 5.14.10-300.fc35.x86_64`

Would it make sense to drop Fedora and use Ubuntu? Seems we don't have to have 22.04 and Vagrant build isn't that unstable (compared to containerd) though.


kolyshkin:
In fact, even 20.04 (on Azure / GHA) switched to kernel 5.15:
```
Ubuntu 20.04.4 LTS
Linux kernel version: 5.15.0-1017-azure
Image Version: 20220828.1
```

so, indeed, we can 
 - drop fedora job
 - switch to ubuntu 22.04 (and either keep or drop ubuntu 20.04)

thaJeztah:
> and either keep or drop ubuntu 20.04

I know 20.04 is currently still a lot more popular than 22.04, so if there's differences that matter for this particular repository, it may be good to keep it for now (I think CI on this repo is fast, so probably should not be a blocker w.r.t. CI duration)

thaJeztah (approved):
LGTM

thaJeztah:
All green now 👍

@kolyshkin LGTY?

samuelkarp (approved):

thaJeztah:
I think this one should be good to go; I'll go ahead and merge, but if there's something we missed, let us know @kolyshkin 👍

PR #128, opened 2023-08-24
mountinfo: linux: use /proc/thread-self/mountinfo

cyphar:
In a Go program with many threads, which goroutine is running on the thread-group leader thread is not something programs can controls. Thus, even if you use runtime.LockOSThread() for threads that have different mount namespaces, it's possible that /proc/self will refer to the "wrong" thread.

The solution is to simply use /proc/thread-self, which will always provide the correct result for the calling thread.

The usage of /proc/self/mountinfo caused isuses for a patch to runc which creates a thread to create id-mapped mounts (which requires joining the container mount namespace).

### DIFF
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -5,15 +5,19 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"strconv"
 	"strings"
+	"sync"
+
+	"golang.org/x/sys/unix"
 )
 
 // GetMountsFromReader retrieves a list of mounts from the
 // reader provided, with an optional filter applied (use nil
 // for no filter). This can be useful in tests or benchmarks
 // that provide fake mountinfo data, or when a source other
-// than /proc/self/mountinfo needs to be read from.
+// than /proc/thread-self/mountinfo needs to be read from.
 //
 // This function is Linux-specific.
 func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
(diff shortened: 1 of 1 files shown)


cyphar:
Ah, this might need `LockOSThread`. Let me do some more testing...

cyphar:
I will need to add some fallback code for pre-3.17 kernels where `/proc/thread-self` didn't exist.

thaJeztah:
> I will need to add some fallback code for pre-3.17 kernels where /proc/thread-self didn't exist.

I guess this was not backported in the RHEL 3.10 kernels? I think RHEL 7 reaches EOL June next year, so ... it's getting closer.

cyphar:
(For future reference, this was motivated by opencontainers/runc#3985.)

cyphar:
Wdyt @kolyshkin?

kolyshkin (approved):
LGTM

mountinfo/mountinfo_linux.go
```diff
@@ -128,7 +131,19 @@ func GetMountsFromReader(r io.Reader, filter FilterFunc) ([]*Info, error) {
...
-	f, err := os.Open("/proc/self/mountinfo")
+	// We need to lock ourselves to the current OS thread in order to make sure
+	// that the thread referenced by /proc/thread-self stays alive until we
+	// finish parsing the file.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	f, err := os.Open("/proc/thread-self/mountinfo")
+	if os.IsNotExist(err) {
+		// On pre-3.17 kernels (such as CentOS 7), we don't have
```
kolyshkin, line 142: The only thing we could add here is a mechanism to remember whether we have the `/proc/thread-self/mountinfo` and skip trying to open it later if we don't have it.

Not sure if this optimization is worth the trouble -- probably not.

cyphar: I have that optimisation in the runc code, I can copy it here if you like?

kolyshkin: Yes please. I guess it makes sense (e.g. for kubelet, which is a long lived daemon).

kolyshkin (approved):
LGTM, thanks!

kolyshkin:
@thaJeztah ptal

kolyshkin:
Merging as this doesn't seem like a breaking change, and we need it in runc (#3985).

cyphar:
Ah, it seems that we need the `/proc/self` fallback that I added to runc in opencontainers/runc#3985 as well (it seems we do get mountinfo in the host rootfs as pid1 in the container...).

PR #129, opened 2023-08-25
Bump CI to use supported go version, fix a single Go 1.20 issue in mount pkg; rm uneeded errorlint annotations

kolyshkin:
1. Bump Go to supported version, other CI components.
2. Bump go to 1.17 in */go.mod.
3. Bump golang.org/x/sys to ~~v0.11.0.~~ v.0.1.0.
4. mount: fix an errorlint issue with Go 1.20+.
5. mount: rm unneeded errorlint annotations (thanks to https://github.com/polyfloyd/go-errorlint/pull/47).
6. sequential: fix a formatting issue (file is not gofumpt'ed)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,12 +4,12 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.17.x, 1.18.x]
+        go-version: [1.17.x, 1.20.x, 1.21.x]
         platform: [ubuntu-20.04, ubuntu-22.04, windows-latest, macos-11]
     runs-on: ${{ matrix.platform }}
     steps:
     - name: Install Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
--- a/mount/go.mod
+++ b/mount/go.mod
@@ -4,5 +4,5 @@ go 1.17
 
 require (
 	github.com/moby/sys/mountinfo v0.6.2
-	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a
+	golang.org/x/sys v0.1.0
 )
--- a/mount/go.mod
+++ b/mount/go.mod
@@ -1,6 +1,6 @@
 module github.com/moby/sys/mount
 
-go 1.16
+go 1.17
 
 require (
 	github.com/moby/sys/mountinfo v0.6.2
--- a/mount/mount_unix.go
+++ b/mount/mount_unix.go
@@ -23,7 +23,7 @@ func Mount(device, target, mType, options string) error {
 // a normal unmount. If target is not a mount point, no error is returned.
 func Unmount(target string) error {
 	err := unix.Unmount(target, mntDetach)
-	if err == nil || err == unix.EINVAL { //nolint:errorlint // unix errors are bare
+	if err == nil || err == unix.EINVAL {
 		// Ignore "not mounted" error here. Note the same error
 		// can be returned if flags are invalid, so this code
 		// assumes that the flags value is always correct.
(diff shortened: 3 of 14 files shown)


.github/workflows/test.yml
```diff
@@ -4,12 +4,12 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.17.x, 1.18.x]
+        go-version: [1.20.x, 1.21.x]
```
thaJeztah, line 7: Given that this is a library, would you be ok with keeping the "oldest" version a bit conservative? I guess we could either go for oldest + latest;

```yaml
        go-version: [1.17.x, 1.21.x]
```

or "oldest" + "current versions"

```yaml
        go-version: [1.17.x, 1.20.x, 1.21.x]
```


I see the `go.mod` of some modules still says `go1.16`, so perhaps that should even be go1.16, or (with Go now no longer considering it "just an indication")  we may have to update the go.mod in all of those modules.

mount/mount_unix.go
```diff
@@ -71,7 +71,7 @@ func RecursiveUnmount(target string) error {
 		if err != nil {
 			if i == lastMount {
 				if suberr != nil {
-					return fmt.Errorf("%w (possible cause: %s)", err, suberr)
+					return fmt.Errorf("%w (possible cause: %w)", err, suberr)
```
thaJeztah, line 74: See my other comment; how much work would it be to continue making this work on older go versions? (I guess we could make the format string conditional)

thaJeztah: Or if we don't really need the `subErr` as an error, we could probably make the linter happy if we use `subErr.Error()`

kolyshkin: > Or if we don't really need the `subErr` as an error, we could probably make the linter happy if we use `subErr.Error()`

That was my initial patch exactly :)

When I figured we have a perfectly working version for older Go releases, and a new one may as well use Go 1.20.

.github/workflows/test.yml
kolyshkin: Done (and switched to Go 1.17 everywhere).

mount/mount_unix.go
kolyshkin: Done (with a TODO added to switch to `%w`).

kolyshkin:
@thaJeztah ptal

mount/go.mod
```diff
@@ -4,5 +4,5 @@ go 1.17
 
 require (
 	github.com/moby/sys/mountinfo v0.6.2
-	golang.org/x/sys v0.0.0-20220520151302-bc2c85ada10a
+	golang.org/x/sys v0.11.0
```
thaJeztah, line 7: if we don't need anything newer, I prefer follow Go module's "minimum version selection", and use the oldest tagged version (v0.1.0), and leave it up to users of this module to pick the version they want

thaJeztah:
@kolyshkin are you ok with picking a "minimum" version for golang.org/x/sys, or did you have a specific reason to _require_ the latest version?

kolyshkin:
> @kolyshkin are you ok with picking a "minimum" version for golang.org/x/sys, or did you have a specific reason to _require_ the latest version?

Yes, I don't think we have a requirement for a specific version here. Will update.

mount/go.mod
kolyshkin: Done, PTAL @thaJeztah

kolyshkin:
@thaJeztah ptal

kolyshkin:
@thaJeztah ptal

thaJeztah (approved):
LGTM! Sorry for the delay, somehow this dropped off my radar 🙈

thaJeztah:
Looks like https://github.com/polyfloyd/go-errorlint/commit/c0e6cac0ec994c118a788505f20fbb20d2b38c11 is part of go-errorlint v1.5.2, which is included in GolangCI-lint v1.59.1; https://github.com/golangci/golangci-lint/releases/tag/v1.59.1

I'll do a follow-up after this to update golangci-lint and to remove that other nolint-comment; also macos-12 runners were deprecated, so will include updates for that as well.

PR #127, opened 2023-06-10
sequential: refactor, and update godoc

thaJeztah:
### sequential: consistently use x/sys/windows for consts

Don't mix-and-mash packages for consts.

### sequential: move error-handling to openFileSequential

Move the error-handling to openFileSequential and update exported functions to use openFileSequential instead of OpenFile().

### sequential: simplify docs for non-Windows implementations

The non-Windows implementations are aliases for the "os" stdlib package equivalents. This patch removes the local documentation, as it was out of date for some, and having the local documentation could give the impression that these functions were anything other than an alias.

Screenshot 2023-06-10 at 12 02 00

### sequential: update docs for Windows-implementation

Update the docs to refer to the "os" stdlib equivalents, and describe the changes compared to those, instead of replicating the os package's documentation.

Screenshot 2023-06-10 at 12 32 45

PR #130, opened 2023-08-25
sequential: simplify aliases

kolyshkin:
This is easier and self-documenting.

A followup to #127.


thaJeztah:
The reason I used a function and not an alias was that pkg.go.dev would show them as "variables" not functions (also not sure if we technically want them to be mutable)

PR #131, opened 2023-08-31
integrate github.com/moby/buildkit/util/appcontext

crazy-max:
- relates to https://github.com/docker/cli/pull/4457

`appcontext` pkg is used across repos and it would be useful to have it as a dedicated module: https://grep.app/search?q=github.com/moby/buildkit/util/appcontext

Specially for `docker/cli` as this is the only package left depending on buildkit: https://github.com/docker/cli/blob/f74f88445f8b23fe4804f1072e5f376cfbe46487/cmd/docker/docker.go#L17

Taken from https://github.com/moby/buildkit/tree/master/util/appcontext commit https://github.com/moby/buildkit/commit/05eb7287534b785822c8cc6c8530f5a6d9609162 with history.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES ?= mountinfo mount sequential signal symlink
+PACKAGES ?= appcontext mountinfo mount sequential signal symlink
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
 	freebsd/amd64 openbsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
--- a/appcontext/appcontext.go
+++ b/appcontext/appcontext.go
@@ -1,0 +1,46 @@
+package appcontext
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"sync"
+
+	"github.com/moby/buildkit/util/bklog"
+)
+
+var appContextCache context.Context
+var appContextOnce sync.Once
+
+// Context returns a static context that reacts to termination signals of the
+// running process. Useful in CLI tools.
+func Context() context.Context {
+	appContextOnce.Do(func() {
+		signals := make(chan os.Signal, 2048)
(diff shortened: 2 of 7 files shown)


thaJeztah:
Thanks! As mentioned on Slack, we probably need to look if docker/cli actually should be using this package; also not 100% sure if moby/sys is the best location for it (but if it still makes sense to have it as a separate module, we can create a repository for it).

neersighted:
Yeah, I don't think this is something we want to do to support the CLI; if the behavior we come up for with the CLI is useful, we can consider moving it here, but I'm not convinced that the logic is generic/fully re-usable yet.

Maybe this makes sense for CLI plugins, but I don't think this is the right module to re-use the code from if we do that.

crazy-max:
Created https://github.com/crazy-max/appcontext that we could move to moby org if that makes sense

thaJeztah:
@crazy-max the CLI no longer uses the package; where are the remaining uses? of it? (considering if it's gonna be used in multiple projects or not)

crazy-max:
Ah right didn't catch https://github.com/docker/cli/pull/4599 removed the dep! All good then!

thaJeztah:
> PS: Does https://github.com/docker/cli/pull/4599 needs a follow-up on buildx ?

Yes, it does! Or at least: the change should be backward-compatible (CLI has fallbacks), but I think @laurazard was planning to open PR's in compose and buildx to update the code to take advantage of the new handling.

I guess; feel free to start making those changes though (and I'm sure @laurazard would be willing to help review them 😄)

laurazard:
Indeed! I have one mostly ready for Compose, was going to start digging into Buildx code this week, happy to do that or review any PRs related to it :)

crazy-max:
> Indeed! I have one mostly ready for Compose, was going to start digging into Buildx code this week, happy to do that or review any PRs related to it :)

Feel free to do so, you seem to have more context (no pun intended) than me :sweat_smile:

PR #135, opened 2023-11-06
mountinfo: linux: add a /proc/self/mountinfo fallback

cyphar:
In very specific cases (that we hit in runc), it is possible to be running in a child pid namespace (where the Go process's pid is 1) but still have a host /proc. While this is not a problem with /proc/thread-self (since the process *is* in the host pid namespace), the /proc/self/task/<tid> fallback doesn't work because gettid(2) returns the wrong thread-id for the host pid namespace.

There isn't much we can do in this case other than use /proc/self, because the simple workarounds (grep NSpid /proc/self/task/*/status) aren't available in pre-3.17 kernels and any more complicated schemes (such as setting the signal mask to something unique and scanning for it) are unworkable for obvious reasons.

Fixes: 12c61a3e3acb ("mountinfo: linux: use /proc/thread-self/mountinfo")

### DIFF
--- a/mountinfo/mountinfo_linux.go
+++ b/mountinfo/mountinfo_linux.go
@@ -156,6 +156,14 @@ func parseMountTable(filter FilterFunc) (_ []*Info, err error) {
 		// /proc/thread-self/ so we need to manually construct
 		// /proc/self/task/<tid>/ as a fallback.
 		f, err = os.Open("/proc/self/task/" + strconv.Itoa(unix.Gettid()) + "/mountinfo")
+		if os.IsNotExist(err) {
+			// If /proc/self/task/... failed, it means that our active pid
+			// namespace doesn't match the pid namespace of the /proc mount. In
+			// this case we just have to make do with /proc/self, since there
+			// is no other way of figuring out our tid in a parent pid
+			// namespace on pre-3.17 kernels.
+			f, err = os.Open("/proc/self/mountinfo")
+		}
 	}
 	if err != nil {
 		return nil, err


cyphar:
This should fix the CI issues in https://github.com/opencontainers/runc/pull/4109. Sorry for the noise :sweat_smile:.

samuelkarp (approved):
LGTM

(because I didn't know offhand if `os.IsNotExist` would work with nil errs: https://go.dev/play/p/Hku5gFHXWtX)

cyphar:
Yup, it seems this does fix the issue. Before:

```
=== RUN   TestUsernsCheckpoint
    checkpoint_test.go:95: unexpected error: unable to start container process: error during container init: error preparing rootfs: open /proc/self/task/1/mountinfo: no such file or directory
--- FAIL: TestUsernsCheckpoint (0.28s)
=== RUN   TestCheckpoint
    checkpoint_test.go:95: unexpected error: unable to start container process: error during container init: error preparing rootfs: open /proc/self/task/1/mountinfo: no such file or directory
--- FAIL: TestCheckpoint (0.21s)
=== RUN   TestExecPS
    exec_test.go:43: |: unable to start container process: error during container init: error preparing rootfs: open /proc/self/task/1/mountinfo: no such file or directory
--- FAIL: TestExecPS (0.15s)
=== RUN   TestUsernsExecPS
    exec_test.go:43: |: unable to start container process: error during container init: error preparing rootfs: open /proc/self/task/1/mountinfo: no such file or directory
--- FAIL: TestUsernsExecPS (0.15s)
```

And now in the test PR (opencontainers/runc#4112), all the tests pass. It is unfortunate that in theory we could produce the wrong data in this one case, but it's only for ancient kernels and (in runc's case at least) we are never running with heterogeneous threads when we are pid1 inside the container.

kolyshkin (approved):
LGTM

Issue #58, opened 2020-12-10
symlink: docs need an update


kolyshkin:
moby/sys/symlink documentation (as seen at e.g. https://pkg.go.dev/github.com/moby/sys/symlink) needs a few updates:

1. Package docs say the package "implements EvalSymlinksInScope ..." while in fact it implements `FollowSymlinkInScope`.

2. Doc for `FollowSymlinksInScope` says it "is a wrapper around evalSymlinksInScope" which is not an adequate description as a user knows nothing about `evalSymlinksInScope`.

kolyshkin:
Also, this is unrelated, but from the first glance it seems that golang's filepath.EvalSymlinks now support `\\?\`-like paths on Windows, which makes moby/sys/symlink.EvalSymlink obsoleted

PR #138, opened 2024-07-03
symlink: touch-up documentation

thaJeztah:
- fixes https://github.com/moby/sys/issues/58

### symlink: touch-up documentation

- Fix references to EvalSymlinksInScope, which is non-exported
- Add relevant links to original implementation / introduction of the code to allow comparing the fork with the original upstream code.
- Move GoDoc from the non-exported evalSymlinksInScope function to the exported FollowSymlinkInScope.
- Remove the README.md, instead using doc.go to provide the package description.

### Before:

Screenshot 2024-07-03 at 10 40 53

### After:

Screenshot 2024-07-03 at 13 05 58

### DIFF
--- a/symlink/doc.go
+++ b/symlink/doc.go
@@ -1,4 +1,11 @@
-// Package symlink implements EvalSymlinksInScope which is an extension of
-// filepath.EvalSymlinks, as well as a Windows long-path aware version of
-// filepath.EvalSymlinks from the Go standard library (https://golang.org/pkg/path/filepath).
+// Package symlink implements [FollowSymlinkInScope] which is an extension
+// of [path/filepath.EvalSymlinks], as well as a Windows long-path aware
+// version of [path/filepath.EvalSymlinks] from the Go standard library.
+//
+// The code from [path/filepath.EvalSymlinks] has been adapted in fs.go.
+// Read the [LICENSE.BSD] file that governs fs.go and [LICENSE.APACHE] for
+// fs_unix_test.go.
+//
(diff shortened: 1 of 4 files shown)


symlink/doc.go
```diff
@@ -1,4 +1,11 @@
-// Package symlink implements EvalSymlinksInScope which is an extension of
-// filepath.EvalSymlinks, as well as a Windows long-path aware version of
-// filepath.EvalSymlinks from the Go standard library (https://golang.org/pkg/path/filepath).
+// Package symlink implements [FollowSymlinkInScope] which is an extension
+// of [path/filepath.EvalSymlinks], as well as a Windows long-path aware
+// version of [path/filepath.EvalSymlinks] from the Go standard library.
+//
+// The code from [path/filepath.EvalSymlinks] has been adapted in fs.go.
+// Read the [LICENSE.BSD] file that governs fs.go and [LICENSE.APACHE] for
+// fs_unix_test.go.
```
kolyshkin, line 7: Can/should we add references to the files we're talking about (i.e. `fs.go` -> `[fs.go]`)?

kolyshkin (approved):
LGTM

symlink/doc.go
thaJeztah: I was considering that initially, but didn't know what the most suitable link for that would be (linking to `main` could cause the link to break potentially, and linking to a versioned one could be confusing when linking from version X to version "old"), so I "somewhat" kept that an exercise to the user for now.

thaJeztah:
I'll bring this one in; will also have a look at possibly doing some patch releases with our latest changes (but need to check how important they are).

PR #140, opened 2024-07-16
integrate libcontainer/userns into moby/sys/user

thaJeztah:
- relates to https://github.com/opencontainers/runc/pull/4331
- relates to https://github.com/opencontainers/runc/issues/3028

---

### integrate libcontainer/userns into moby/sys/user

This integrates the userns package from libcontainer (runc) at commit;
3778ae603c706494fd1e2c2faf83b406e38d687d (https://github.com/opencontainers/runc/commit/3778ae603c706494fd1e2c2faf83b406e38d687d)

Using the following steps:

```bash
brew install git-filter-repo

cd ~/Projects
git clone https://github.com/opencontainers/runc.git runc_temp
cd runc_temp

# commit taken from
git rev-parse --verify HEAD
3778ae603c706494fd1e2c2faf83b406e38d687d

# remove all code, except for 'libcontainer/userns, and rename to /user/userns
git filter-repo \
  --path 'libcontainer/userns/' \
  --path-rename libcontainer/userns:user/userns

git checkout -b integrate_libcontainer_userns

git remote add libcontainer_userns ~/Projects/runc_temp
git fetch libcontainer_userns

# merge the upstream code
git merge --allow-unrelated-histories --signoff -S libcontainer_userns/main
```


Makefile
```diff
@@ -16,14 +16,18 @@ clean:
 test: test-local
 	set -eu; \
 	for p in $(PACKAGES); do \
-		(cd $$p; go test $(RUN_VIA_SUDO) -v .); \
+		if $p = user && go version | grep -qv go1.17; then \
```
AkihiroSuda, line 19: No need to support Go <= 1.20

thaJeztah: The other modules still have go1.17 as minimum version; I didn't want to change that as part of this PR, but we can revisit that in a follow up

AkihiroSuda (approved):

user/userns/userns_linux.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"os"
+	"sync"
+)
+
+var inUserNS = sync.OnceValue(runningInUserNS)
+
+// runningInUserNS detects whether we are currently running in a user namespace.
+//
+// Originally copied from https://github.com/lxc/incus/blob/e45085dd42f826b3c8c3228e9733c0b6f998eafe/shared/util.go#L678-L700.
```
kolyshkin, line 14: nit: we should probably refer to runc, too.

kolyshkin:
It looks like git history here begins when the code was moved in runc, and contains some drastic changes (like additon and removal of a big portion of C code). IOW it's not very useful (I tried to answer simple questions like "why do we use `bytes.Buffer`?" or "who added the constant `4294967295`?" using `git blame` and `git log`, and it's not possible to do so).

Given that git history serves a main purpose of figuring out why (or when, or by whom) things were written the way they are written, and that this git history is:

1. quite lengthy (esp. compared to the amount of code added), yet
2. doesn't work for the above purpose,

I'm not sure it makes sense to retain it.

I think it's better to just copy and commit the code @ specific commit with a clean git history, and refer to older runc history in the docs (like we did when it was copied from lxc).

user/userns/userns_linux.go
thaJeztah: Added a mention of runc (both here, and in the package description)

thaJeztah:
@kolyshkin I gave this another go, and used more specific filters so that the unrelated commits don't get included. History now contains all history since the code was moved, but the merge-commit contains a link to the commits that moved it.

Let me know if this looks better 🤞

thaJeztah:
@kolyshkin ptal 🤗

cyphar (approved):
LGTM

thaJeztah:
oh! need to rebase after https://github.com/moby/sys/pull/142 was merged

laurazard (approved):
LGTM

tianon:
I'm a little late to the party (sorry 😭), but I'm actually a bit confused by this.  As far as I can tell, the only thing the "user" and "userna" packages have in common is the name, so this seems like an odd grouping IMO

tianon:
I'm also missing the justification for the bump to go1.21? https://github.com/moby/sys/commit/5cd502ca1e109c963ac486b3e8f84ae73ff1a74f#r144637064

(and in fact, https://github.com/moby/sys/pull/140#discussion_r1680259311 seems to directly contradict doing so :thinking:)

thaJeztah:
@tianon Arf, sorry for ghosting you; I saw your comment on my phone, and wanted to bring it up in one of the calls, but I didn't make both of them, then forgot.

> I'm a little late to the party (sorry 😭), but I'm actually a bit confused by this. As far as I can tell, the only thing the "user" and "userns" packages have in common is the name, so this seems like an odd grouping IMO

That, hm, is a valid point. I think I initially put it as a separate module, but then (maybe incorrectly?) recalled that there was a discussion to move this, as well as possibly the `idtools` bit into the same module. But maybe I my memory fails me here (it's been leaky a bit recently).

FWIW; to my knowledge none of the current consumers (moby, buildkit, containerd) have included this change in a release, so there's still a potential to change this to a separate module if we think that's better

> I'm also missing the justification for the bump to go1.21? [5cd502c#r144637064](https://github.com/moby/sys/commit/5cd502ca1e109c963ac486b3e8f84ae73ff1a74f#r144637064)

The code in `userns` uses [`sync.OnceFunc`](https://pkg.go.dev/sync#OnceValue), which requires go1.21

> (and in fact, [#140 (comment)](https://github.com/moby/sys/pull/140#discussion_r1680259311) seems to directly contradict doing so 🤔)

Ah, right, so the comment was about the added check in `make test`; GitHub actions runs on go1.18, go1.21.x and go1.22.x; the lines added are to skip the `user` module on go1.18, but still test the other modules (mount, mountinfo, sequential, signal, symlink) on that version.

tianon:
The context for why I even care is that I'm a downstream user of `github.com/moby/sys/user` who will never touch `userns`, so this change has me strongly considering the [Go Proverb](https://go-proverbs.github.io/) "A little copying is better than a little dependency." (but "a little dependency" is part of why `github.com/moby/sys/user` was created / exists in the first place) :sweat_smile: :see_no_evil:

tianon:
So that's the long way of saying I'm very much in favor of a `git mv user/userns userns`

thaJeztah:
I opened a PR to move this to a separate module;

- https://github.com/moby/sys/pull/145

PR #145, opened 2024-08-06
user: move userns package to separate module, and retract v0.2.0

thaJeztah:
Relates to:

- https://github.com/moby/sys/pull/140
- https://github.com/opencontainers/runc/issues/3028

Related downstream PRs;

- (merged, but not released); https://github.com/moby/buildkit/pull/5189
- (merged, but not released); https://github.com/moby/moby/pull/48170 / https://github.com/moby/moby/pull/48238
- (merged, but not released); https://github.com/containerd/containerd/pull/10473
- (not yet merged); https://github.com/containerd/cgroups/pull/346
- (not yet merged); https://github.com/opencontainers/runc/pull/4350

commit 86870e7b5ae771ec567bf0519663b95345ba59e3 integrated the userns package into the github.com/moby/sys/user module, which was included in the v0.2.0 version of the module.

Upon further discussion with maintainers, this may not have been a good choice; the userns package is related to user-namespaces (uid/gid-mapping), and while there are some tangential relations with "user", we shouldn't conflate these concepts by putting both into the same module.

Some downstream projects (containerd, moby, containerd/cgroups) already accepted patches to switch to the package that's part of the moby/sys/user module, but none of those patches made it into a release.

This patch:

- moves the userns package to a separate module
- retracts the moby/sys/user v0.2.0 release
- downgrades the minimum go version update for the moby/sys/user module to go1.17. Note that CI is no longer testing go1.17, but go1.18 as minimum.


thaJeztah:
@tianon @AkihiroSuda @corhere PTAL

tianon (approved):
:heart: :heart: :heart:

Makefile
```diff
@@ -16,7 +16,7 @@ clean:
 test: test-local
 	set -eu; \
 	for p in $(PACKAGES); do \
-		if $p = user && go version | grep -qv go1.18; then \
+		if $p = userns && go version | grep -qv go1.18; then \
```
tianon, line 19: Wait, this syntax is funky -- `if $p = userns` is going to run `$p` as an executable with arguments `= userns` and since `$p` is a directory, will probably _always_ be false and thus this block never runs?

tianon:
```suggestion
		if [ "$p" = userns ] && go version | grep -qv go1.18; then \
```

(probably meant something like this here and in the other loops in this file)

tianon: Also, we _only_ run the tests on go1.18?  That seems strange? (I'm likely missing some context)

corhere (approved):

thaJeztah:
Wah, sorry for the repeated push; I was fixing up the conditional skip; I gave up on trying to cramp it into the makefile as the combination of conditionals, subshells and windows became too fiddly; instead just made the list of modules to test conditional in the GitHub action

@tianon PTAL if you're more happy with this approach 😅

.github/workflows/test.yml
```diff
@@ -14,6 +14,9 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
       uses: actions/checkout@v4
+    - if: ${{ matrix.go-version == '1.18.x' }}
+      run: |
+        echo 'PACKAGES="mountinfo mount sequential signal symlink user"' >> $GITHUB_ENV
```
tianon, line 19: This is the same list as `Makefile:1`, but minus `userns`, right?  That's somewhat clear while reviewing this diff while they're close together, but will no longer be clear the minute we merge this. :sweat_smile:  How do we make sure we remember to update this list in the future?  (and know how to update it correctly?)

thaJeztah: Yeah, no good solution for that, honestly. Unless we probably introduce more complexity.

Overall, the risk is low;

- if we don't update, the default is still all packages from the Makefile
- if a new module is added that doesn't support go1.18, that module will thus fail
- if we add a new module that should support go1.18, but we forgot to update; that module is skipped in the go1.18 tests (but still tested on current versions)


I'd consider only the last one slightly problematic, but perhaps a calculated risk, given that adding more modules should be a rare event 🙈

thaJeztah: erm: middle bullet is obviously wrong; so only 1st and last apply 🤦‍♂️

.github/workflows/test.yml
```diff
@@ -14,6 +14,11 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
       uses: actions/checkout@v4
+    - if: ${{ matrix.go-version == '1.18.x' }}
+      run: |
+        # This corresponds with the list in Makefile:1, but omits the "userns"
+        # module, which requires go1.21 as minimum.
```
thaJeztah, line 20: I added a comment here to relate it to the variable in the Makefile

Makefile
```diff
@@ -1,4 +1,4 @@
-PACKAGES ?= mountinfo mount sequential signal symlink user
+PACKAGES ?= mountinfo mount sequential signal symlink user userns # IMPORTANT: when updating this list, also update the conditional one in .github/workflows/test.yml
```
thaJeztah, line 1: And added a comment here as a reminder to check the one in the GitHub actions workflow when updating.

tianon (approved):
:heart_hands:

tianon:
Changes since Cory's review are pretty minimal overall, so I'm going to go ahead and merge, but of course we can always open a follow-up if doing so was premature. :heart:

PR #149, opened 2024-09-06
Add capability package

kolyshkin:
This integrates github.com/kolyshkin/capability (which itself is a fork
of github.com/syndtr/gocapability) to github.com/moby/sys/capability,
as suggested in https://github.com/opencontainers/runc/pull/4358#discussion_r1699684243

Some of github.com/syndtr/gocapability users are (in an alphabetical order):
    
 - https://github.com/canonical/lxd
 - https://github.com/containers/buildah
 - https://github.com/containers/podman
 - https://github.com/containers/skopeo
 - https://github.com/google/gvisor
 - https://github.com/hashicorp/nomad
 - https://github.com/linuxkit/linuxkit
 - https://github.com/opencontainers/runc
 - https://github.com/opencontainers/runtime-tools
 - https://github.com/slimtoolkit/slim
 
 The way it was done is:
 
```
# 1. Create a copy of a local repo: 
git clone --no-local capability capability-for-merge
 
# 2. Use git-filter-branch to:
#  - move contents to a subdirectory
#  - remove useless merge commits
#  - remote .github files
#  - add 'capability: ' prefix to commit messages
cd capability-for-merge
git filter-repo --prune-degenerate always --path-rename :capability/ \
    --path-match .github --invert-paths \
    --message-callback '
  prefix = b"capability: "
  if not message.startswith(prefix):
    message=prefix+message
  return message
'

# 3. Merge the above repo to moby/sys
cd ~/git/moby/sys
git checkout -b add-capability
git remote add cap-merge ../../capability-for-merge
git fetch cap-merge
git merge --allow-unrelated-histories --signoff -S cap-merge/main
```


kolyshkin:
I guess we're OK with the DCO check failing, it would be unwise to rewrite 12 year old commits.

thaJeztah:
I marked the DCO to pass; for posterity; these are the commits that failed the DCO;


Screenshot 2024-09-09 at 12 45 18

capability/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright 2013 Suryandaru Triandana <syndtr@gmail.com>
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
```
thaJeztah, line 6: I'm curious what the right approach is for this license file; currently this means that `golang.org/moby/sys/capability` will be licensed as MIT, but the LICENSE at the top of the repository is Apache2.


I know when vendoring other modules from this repository, `go mod` copies the LICENSE from the top-level, so I wonder what it does for this module (will it copy _both_? will it only use the MIT license?).

Reading these two StackOverflow answers;

- https://opensource.stackexchange.com/a/7391
- https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project

---

- At least the original license and copyright must be retained (see below)
- Possibly as a secondary file (?)
- Possibly included in a NOTICE file?

thaJeztah (changes requested):
Thanks! Forgot to submit some comments I had

capability/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/kolyshkin/capability
+
+go 1.21
```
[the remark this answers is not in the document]
kolyshkin: I _think_ there's nothing that requires Go newer than 1.18. Set to 1.18, let's see if it works.

capability/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/kolyshkin/capability
```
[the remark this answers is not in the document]
kolyshkin: fixed, thanks!

capability/go.mod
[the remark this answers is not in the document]
kolyshkin: I was terribly wrong. The code is now using `sync.OnceValues` which was added in Go 1.21.

capability/LICENSE
kolyshkin: I am now testing this PR in https://github.com/opencontainers/runc/pull/4358, and from what I see, `go mod vendor` uses top-level LICENSE file for e.g. moby/sys/mount and other packages, and it uses directory-level LICENSE for `moby/sys/capability`. In other words, it works as it should.

I don't see any problem with this layout -- a LICENSE file in a subdirectory overwrites the upper level LICENSE. If they are compatible (as in our case), all is fine and dandy.

Disclaimer: IANAL.

thaJeztah: Yeah, I'm somewhat looking from a "module" perspective, because it may become somewhat ambiguous; for the GitHub repo, the license on the repo-level is Apache 2, with a subdirectory having the MIT license.

But from a module perspective, normally it would copy that license; in this case it only has the MIT license. And perhaps that's OK (so, the `capabilities` module having a different license than the other modules), but was wondering if we wanted to re-license the new module to use Apache 2.

thaJeztah: ☝️ s/MIT/BSD 2-clause/ (I was confused)

tianon (approved):
BSD-2-clause is different from Apache 2, but strictly _less_ restrictive, so is probably OK.  Trying to relicense is going to be a huge pain. :+1:

This seems fine to me (currently discussing some minor suggested changes in the Moby maintainers meeting, but I don't think there's any strong opposition overall).

thaJeztah (approved):
LGTM

capability/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright 2013 Suryandaru Triandana <syndtr@gmail.com>
```
thaJeztah, line 1: We can add a line above this "Copyright the authors" (which would apply to new contributions)

thaJeztah:
@AkihiroSuda PTAL if this LGTY

capability/LICENSE
kolyshkin: > golang.org/moby/sys/capability will be licensed as MIT

For the sake of those reading this thread. This is actually a BSD 2-clause license, not the MIT license.
See https://opensource.org/license/bsd-2-clause

capability/LICENSE
kolyshkin: Done; PTAL.

capability/CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
+# Changelog
+This file documents all notable changes made to this project since the initial fork.
```
AkihiroSuda, line 2: The fork history should be explained

capability/CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.1.0] - 2024-07-31
+
+This is an initial release since the fork.
+
+### Breaking changes
+
+ * The `CAP_LAST_CAP` variable is removed; users need to modify the code to
+   use [LastCap] to get the value. (#6)
```
AkihiroSuda, line 14: PR numbers are no longer valid

capability/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright 2023 The Capability Authors.
```
AkihiroSuda, line 1: And Moby authors

thaJeztah:
Also /cc @dims (FYI - you probably also have an interest in this)

capability/LICENSE
thaJeztah: We briefly discussed options here; we could even choose to make it just "The Authors", which is the most generic form; ultimately, copyright of any code change belongs to the author (what project that code change was made in, doesn't really matter there).

I guess the only benefit of mentioning the project is "where did it originate from" in cases where the code is forked by others.

dims:
of course! this is a welcome development. thanks for the poke @thaJeztah

capability/CHANGELOG.md
kolyshkin: I have added a link to the original repo here. The move to moby/sys will be explained in 0.2.0 changelog entry.

capability/CHANGELOG.md
kolyshkin: Fixed.

capability/LICENSE
kolyshkin: I think "the Capability Authors" covers everyone who contributes to this repository.

kolyshkin:
Just pushed a new version, mostly addressing @AkihiroSuda's comments, i.e.
 - fixed CHANGELOG.md to clarify the fork source and fix links to PRs.
But also:
 - added [the missing commit from the kolyshkin/capability repo](https://github.com/kolyshkin/capability/commit/d1331df7ef9b2f2e21f5ec73fd7a051a7cc9e15e) which was lost before;
 - fixed a small issue in a changelog (go version requirement was wrong).
 
All that (except for the missing commit addition) are added as the last commits, so you are re-reviewing, you can only check the last commits.

AkihiroSuda (approved):

kolyshkin:
I think we can address any other concerns post-merge.

thaJeztah:
Thanks 🙏

PR #151, opened 2024-09-20
capability: doc improvements

kolyshkin:
Assorted doc improvements:
* doc: add cross-references;
* doc: use tabs in code examples;
* README: add link to changelog;
* CHANGELOG: link to 0.2.0 release.

### DIFF
--- a/capability/CHANGELOG.md
+++ b/capability/CHANGELOG.md
@@ -5,7 +5,7 @@ from https://github.com/syndtr/gocapability/commit/42c35b4376354fd5.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.2.0 - 2024-09-16
+## [0.2.0] - 2024-09-16
 
 This is the first release after the move to a new home in
 github.com/moby/sys/capability.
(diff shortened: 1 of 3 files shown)


capability/CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ This is an initial release since the fork.
 <!-- Minor releases. -->
 [0.1.1]: https://github.com/kolyshkin/capability/compare/v0.1.0...v0.1.1
 [0.1.0]: https://github.com/kolyshkin/capability/compare/42c35b4376354fd5...v0.1.0
+[0.2.0]: https://github.com/moby/sys/releases/tag/capability%2Fv0.2.0
```
tianon, line 60: Minor nit, but shouldn't this be added to the top of this list instead of the bottom? (they're otherwise in descending order, which matches the order of their usage above)

tianon (approved):
Left a minor comment, but LGTM either way

capability/CHANGELOG.md
kolyshkin: Right; fixed.

kolyshkin:
Merging with 1 reviewer only as it's a minor improvement which should not affect any code, and we'll be able to see the results of this in go.pkg.dev once merged (to assess if all the changes are correct and, if anything is wrong, fix before the release).

kolyshkin:
Oh well, we hit this issue: https://github.com/golang/go/issues/62293

PR #152, opened 2024-09-20
capability: add LastCap stub for non-Linux

kolyshkin:
So that the code which uses capability.LastCap can be compiled on
non-Linux platforms.

Found the issue while working on https://github.com/containers/common/pull/2167


tianon (approved):

kolyshkin:
Merging with one approver only as this is a simple (almost trivial) change.

PR #153, opened 2024-09-20
capability: add ListKnown, ListSupported; deprecate List

kolyshkin:
Apparently, most users of capability.List wants the list of supported
capabilities (i.e. they go on to exclude capabilities above the last
known one).
    
Let's provide `ListSupported` to such users.
   
Also, provide `ListKnown`, and deprecate `List`.

Finally, amend LastCap documentation with a link to ListSupported,
as this is what most users are using it for.

~~Draft until #152 is merged; TODO: amend `LastCap` doc to mention `ListSupported`.~~


tianon (approved):
Nice :eyes:

kolyshkin:
@thaJeztah PTAL

kolyshkin:
Rebased

capability/enum.go
```diff
@@ -301,3 +303,30 @@ const (
...
+// ListKnown returns the list of all capabilities known to the package.
+func ListKnown() []Cap {
+	return list()
+}
+
+// ListSupported retuns the list of all capabilities known to the package,
+// except those that are not supported by the currently running Linux kernel.
+//
+// Returns nil on platforms other than Linux.
+func ListSupported() []Cap {
```
kolyshkin, line 323: I'm not entirely sure whether we should switch here to returning `[]Cap, error`.

On one hand, `nil` by itself is legit, and a function which only returns a slice can be used in a loop directly:

```go
for cap := range ListSupported() {
  ...
}
```

And, if needed, an error can be obtained from `LastCap`:

```go
list := ListSupported()
if list == nil {
   _, err := LastCap()
}
```
(and we can document that LastCap is to be used to obtain an error).

On the other hand, it's easier just to return `nil, err`. The downside is it's not possible to use two returns directly in a `for` loop, you have to use an intermediate variable:

```go
list, _ := ListSupported()
for cap := range list {
   .... 
}
```

kolyshkin: Let's add the error return, it's easier that way.

PR #154, opened 2024-09-20
capability: prepare for v0.3.0 release 

kolyshkin:
Add 0.3.0 changelog and release date.


tianon (approved):
LGTM (although https://github.com/moby/sys/pull/151#discussion_r1769355818 applies here too -- alternatively, if they're easier to maintain in append-order instead of prepend-order, maybe we swap 0.1.1 and 0.1.0 and call it a day?)

kolyshkin:
> LGTM (although [#151 (comment)](https://github.com/moby/sys/pull/151#discussion_r1769355818) applies here too -- alternatively, if they're easier to maintain in append-order instead of prepend-order, maybe we swap 0.1.1 and 0.1.0 and call it a day?)

Thanks, fixed here, too.

PR #157, opened 2024-09-21
ci: fix not honoring PACKAGES setting in gha Windows runners

### LINKED ISSUE #156: ci: PACKAGES hack doesn't work on Windows
Apparently the hack introduced by #145 (using custom PACKAGES list for older Go versions) is not working and never worked on Windows.

Here is an example of a run which has the full list of PACKAGES on Windows:

https://github.com/moby/sys/actions/runs/10298746983/job/28504536200

You can see it works fine on other platforms (Linux and Mac OS X).

kolyshkin:
Fixes #156.


thaJeztah (approved):
LGTM

PR #158, opened 2024-09-24
ci: add unparam linter

kolyshkin:


.golangci.yml
```diff
@@ -1,5 +1,6 @@
 linters:
   enable:
+    - unparam
     - unconvert
     - errorlint
```
thaJeztah, line 5: We should sort this list; helps preventing accidental duplicates and makes it easier to find which ones are enabled

mountinfo/mounted_linux_test.go
```diff
@@ -15,252 +15,247 @@ import (
...
+func tMount(t *testing.T, src, dst, fstype string, flags uintptr) {
 	if os.Getuid() != 0 {
 		t.Skip("root required for mounting")
 	}
 
-	err := unix.Mount(src, dst, fstype, flags, options)
+	err := unix.Mount(src, dst, fstype, flags, "")
 	if err != nil {
-		return &os.PathError{Path: dst, Op: "mount", Err: err}
+		t.Helper()
```
thaJeztah, line 27: Can you put this one at the start of the function? I think it's the more natural place to put it; also makes sure that the `t.Skip()` shows the right line-number of the test, not the helper.

mountinfo/mounted_linux_test.go
```diff
@@ -15,252 +15,247 @@ import (
...
 	t.Cleanup(func() {
 		if err := unix.Unmount(dst, unix.MNT_DETACH); err != nil {
+			t.Helper()
 			t.Errorf("cleanup: unmount %q failed: %v", dst, err)
 		}
 	})
-	return nil
+}
+
+func tCreateFile(t *testing.T, dir string) string {
```
thaJeztah, line 38: This should probably be a `t.Helper()` as well?

mountinfo/mounted_linux_test.go
kolyshkin: Thanks! Done.

mountinfo/mounted_linux_test.go
kolyshkin: Right.

.golangci.yml
kolyshkin: done

kolyshkin:
Fixed review comments; PTAL @thaJeztah

thaJeztah (approved):
LGTM

PR #159, opened 2024-09-26
ci: add codespell

kolyshkin:
Fix typos, rename some variables etc, add codespell job to CI.

Please see individual commits for details.

~~_Currently a draft pending #162 merge._~~

### DIFF
--- a/signal/signal.go
+++ b/signal/signal.go
@@ -12,14 +12,12 @@ import (
 )
 
 // CatchAll catches all signals and relays them to the specified channel.
-// SIGURG is not handled, as it's used by the Go runtime to support
-// preemptable system calls.
+// SIGURG is ignored, as it is used by the Go runtime for internal purposes
+// (see https://github.com/golang/go/issues/24543).
 func CatchAll(sigc chan os.Signal) {
 	var handledSigs []os.Signal
 	for n, s := range SignalMap {
 		if n == "URG" {
-			// Do not handle SIGURG, as in go1.14+, the go runtime issues
-			// SIGURG as an interrupt to support preemptable system calls on Linux.
 			continue
 		}
 		handledSigs = append(handledSigs, s)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,3 +45,9 @@ jobs:
         uname -a
         make test
 
+  codespell:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
+    - run: pip install --break-system-packages codespell==v2.3.0
+    - run: codespell
--- a/capability/.codespellrc
+++ b/capability/.codespellrc
@@ -1,3 +1,0 @@
-[codespell]
-skip = ./.git
-ignore-words-list = nd
--- a/capability/enum.go
+++ b/capability/enum.go
@@ -316,7 +316,7 @@ func ListKnown() []Cap {
 	return list()
 }
 
-// ListSupported retuns the list of all capabilities known to the package,
+// ListSupported returns the list of all capabilities known to the package,
 // except those that are not supported by the currently running Linux kernel.
 func ListSupported() ([]Cap, error) {
 	last, err := LastCap()
--- a/symlink/fs.go
+++ b/symlink/fs.go
@@ -110,15 +110,15 @@ func evalSymlinksInScope(path, root string) (string, error) {
 		// into "/b/../c" which then gets filepath.Cleaned into "/c" and then
 		// root gets prepended and we Clean again (to remove any trailing slash
 		// if the first Clean gave us just "/")
-		cleanP := filepath.Clean(string(filepath.Separator) + b.String() + p)
-		if isDriveOrRoot(cleanP) {
+		pClean := filepath.Clean(string(filepath.Separator) + b.String() + p)
+		if isDriveOrRoot(pClean) {
 			// never Lstat "/" itself, or drive letters on Windows
 			b.Reset()
 			continue
 		}
-		fullP := filepath.Clean(root + cleanP)
+		pFull := filepath.Clean(root + pClean)
 
-		fi, err := os.Lstat(fullP)
+		fi, err := os.Lstat(pFull)
 		if os.IsNotExist(err) {
 			// if p does not exist, accept it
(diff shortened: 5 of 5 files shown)


capability/capability_test.go
```diff
@@ -48,17 +62,7 @@ func TestLastCap(t *testing.T) {
 
 func TestListSupported(t *testing.T) {
 	list, err := ListSupported()
```
thaJeztah, line 64: Looks like `list` here is shadowing `list()` in enum_gen.go; perhaps `supported` would work

capability/capability_test.go
```diff
@@ -21,17 +21,31 @@ const (
...
 )
 
-func TestLastCap(t *testing.T) {
-	last, err := LastCap()
+// checkErrLinux does two things. First, it fails the test if err is:
+//   - not nil on Linux;
+//   - nil on other platforms.
+//
+// Returns true for Linux, false for other platforms.
+func checkErrLinux(t testing.TB, err error) bool {
```
thaJeztah, line 29: Hm.. so I'm a bit on the fence here; it feels like the helper abstraction only saves a few lines of code, but at the cost of now having to explain what it does (which is not immediately obvious by _only_ looking at the calls in the `TestXXX` functions), and now the `t.Fatal()` has to explain that it was non-linux because the test-failure would now show the line-number of the call-site, which is either Linux or non-Linux.

At least for me, it's _more_ cognitive load than the few lines saved.

We're also not checking for a specific error, which is something we could consider (looks like the total number of lines is even shorter than with the helper).

Finally, I noticed we're using `t.Fatal` in various places in this file; looking at most places, we could use `t.Error` instead, and not fail early; doing so _could_ still provide more context about the failing test, which could help estimate what the issue is.


Here's a quick draft what it _could_ look like without the helper;

```go
func TestLastCap(t *testing.T) {
	last, err := LastCap()
	if runtime.GOOS != "linux" {
		if !errors.Is(err, errNotSup) {
			t.Errorf("expected %v, got %v", errNotSup, err)
		}
		return
	}

	// Sanity checks (Linux only).
	if err != nil {
		t.Errorf("expected nil error, got %v", err)
	}
	if last < minLastCap {
		t.Errorf("LastCap returned %d (%s), expected >= %d (%s)", last, last, minLastCap, minLastCap)
	}
	if last > maxLastCap {
		t.Errorf("LastCap returned %d, expected <= %d (%s). Package needs to be updated.", last, maxLastCap, maxLastCap)
	}
}

func TestListSupported(t *testing.T) {
	supported, err := ListSupported()
	if runtime.GOOS != "linux" {
		if !errors.Is(err, errNotSup) {
			t.Errorf("expected %v, got %v", errNotSup, err)
		}
		return
	}

	// Sanity check (Linux only).
	if err != nil {
		t.Errorf("expected nil error, got %v", err)
	}
	t.Logf("got +%v (len %d)", supported, len(supported))
	minLen := int(minLastCap) + 1
	if len(supported) < minLen {
		t.Errorf("result is too short (got %d, want %d): +%v", len(supported), minLen, supported)
	}
}
```

capability/capability_test.go
```diff
@@ -21,17 +21,31 @@ const (
...
 	switch runtime.GOOS {
 	case "linux":
 		if err != nil {
+			t.Helper()
```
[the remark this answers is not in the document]
kolyshkin: I have caught this patter from go sources. In many places they do it this way (supposedly because `t.Helper()` is costly, and if we can avoid it we should -- but I haven't checked for sure, just thought that they know what they are doing 😁

capability/capability_test.go
kolyshkin: This is not part of this PR, this code is from #153.

We can rename this package capability_test :) to make this shadowing go away 🕺🏻

signal/signal.go
```diff
@@ -13,13 +13,11 @@ import (
 
 // CatchAll catches all signals and relays them to the specified channel.
 // SIGURG is not handled, as it's used by the Go runtime to support
-// preemptable system calls.
+// preemptible system calls.
```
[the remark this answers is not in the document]
kolyshkin: I think we can rephrase this without using any of these two words.

Done

capability/capability_test.go
kolyshkin: 1. There's also a third test case which uses the same helper, and I expect more will be added. Also, we might have to add "android" next to "linux", and without a helper we'll have to do it in multiple places, but I'm with you.

2. Re: using `errNotSup` -- it's an internal ID, not exposed through the API, so I don't think it makes sense to check for it.

3. Re: this code:
```go
	// Sanity check (Linux only).
	if err != nil {
		t.Errorf("expected nil error, got %v", err)
	}
	t.Logf("got +%v (len %d)", supported, len(supported))
	minLen := int(minLastCap) + 1
	if len(supported) < minLen {
```

The first `t.Errorf` should be `t.Fatalf` because if List returned an error, we don't want to check the list length. Same in another test.

4. I used `want, got` instead of `expected, got` as I picked it up from Go sources, it's kind of a standard in there and they don't like other working. I find it short(er) and convenient.

5. `runtime.GOOS` in an error message is a subtle reminder why we expect non-nil error (because "Windows" etc).

6. I feel these changes are well beyond the scope of adding codespell, so they deserve a separate PR. Removed non-trivial changes from this PR, will open a new one soon.

capability/capability_test.go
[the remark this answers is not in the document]
kolyshkin: This code is removed, and this patch is now moved to #162.

capability/capability_test.go
kolyshkin: Code moved to #162; in there I actually followed my own idea of renaming the test package.

capability/capability_test.go
kolyshkin: @thaJeztah ptal #162

kolyshkin:
Rebased; no longer a draft.

tianon (approved):

kolyshkin:
@thaJeztah PTAL

kolyshkin:
Merging with a single LGTM since the changes are minor; if anything comes up, lets fix is post merge.

PR #162, opened 2024-09-27
capability: use strings.CutPrefix; test improvements

kolyshkin:
See individual commits for details. Originally part of #159.

### DIFF
--- a/capability/capability_test.go
+++ b/capability/capability_test.go
@@ -2,11 +2,13 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package capability
+package capability_test
 
 import (
 	"runtime"
 	"testing"
+
+	. "github.com/moby/sys/capability"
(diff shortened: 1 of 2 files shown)


capability/capability_test.go
```diff
@@ -24,47 +26,43 @@ const (
...
-	case "linux":
-		if err != nil {
-			t.Fatal(err)
-		}
 	default:
 		if err == nil {
 			t.Fatal(runtime.GOOS, ": want error, got nil")
 		}
 		return
+	case "linux":
```
kolyshkin, line 34: I like it that you can have `default` as the first case in switch; it improves readability in here (otherwise I would have to add a comment saying "Below is linux-only stuff").

kolyshkin:
@thaJeztah @tianon PTAL :pray:

I have some more changes that depend on these.

tianon (approved):

PR #164, opened 2024-09-27
use pr_CAP_AMBIENT_CLEAR_ALL to reduce pr_CAP_AMBIENT_LOWER syscall

### LINKED ISSUE #163: capability: reduce prctl syscall when setting ambient cap set
Even though the ambient cap set is empty, it will always call `PR_CAP_AMBIENT_LOWER` for each support caps, in most machines, it will call 40 times.
We can use `PR_CAP_AMBIENT_CLEAR_ALL` to avoid this big for loop to reduce the `prctl` syscall times.

Please see:
https://github.com/moby/sys/blob/46235e82ffae28e015f1e5e1e3a5c1d51603ad55/capability/capability_linux.go#L366-L373

lifubang:
When we want to set ambient capability for current process/thread, we can clear it first, and set them later one by one.
We don't need to clear all supported but unused capabilities one by one.

Close #163


capability/capability_linux.go
```diff
@@ -364,19 +364,22 @@ func (c *capsV3) Apply(kind CapType) (err error) {
...
 
 	if kind&AMBS == AMBS {
+		err = prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0)
+		if err != nil && err != syscall.EINVAL { //nolint:errorlint // Errors from syscall are bare.
+			// Ignore EINVAL as not supported on kernels before 4.3
+			return
+		}
 		for i := Cap(0); i <= last; i++ {
-			action := pr_CAP_AMBIENT_LOWER
 			if c.Get(AMBIENT, i) {
```
kolyshkin, line 373: nit: can we please do
```go
if ! c.Get(AMBIENT, i) {
	continue
}
err = prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_RAISE, uintptr(i), 0, 0)
...
```
instead? This reduces indentation and improves readability a bit.

kolyshkin:
This is definitely an improvement, thank!

Can we have a test case added?

capability/capability_test.go
```diff
@@ -68,3 +78,56 @@ func TestListSupported(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	// Restore the cap set data from current process
+	if err = pid.Load(); err != nil {
+		t.Fatal(err)
+	}
+	for _, cap := range capAmbient {
+		if !pid.Get(AMBIENT, cap) {
+			t.Fatalf("Can't get ambient cap(%d) from current process.\n", cap)
```
kolyshkin, line 113: nit: no need for `\n` here.

capability/capability_test.go
```diff
@@ -68,3 +78,56 @@ func TestListSupported(t *testing.T) {
...
+	pid.Set(AMBIENT, capAmbient[0])
+	if err = pid.Apply(CAPS | BOUNDING | AMBIENT); err != nil {
+		t.Fatal(err)
+	}
+
+	if err = pid.Load(); err != nil {
+		t.Fatal(err)
+	}
+	if !pid.Get(AMBIENT, capAmbient[0]) {
+		t.Fatalf("Can't get ambient cap(%d) from current process.\n", capAmbient[0])
```
kolyshkin, line 128: ditto

capability/capability_test.go
```diff
@@ -68,3 +78,56 @@ func TestListSupported(t *testing.T) {
...
+	}
+	if pid.Get(AMBIENT, capAmbient[1]) {
+		t.Fatalf("Should not have the ambient cap(%d) in current process.\n", capAmbient[1])
```
kolyshkin, line 131: ditto

capability/capability_test.go
kolyshkin: A more proper error here would be "expected ambient cap(%d) to be set bit it's not".

capability/capability_test.go
kolyshkin: A more proper error here would be "expected ambient cap(%d) to be set bit it's not".

kolyshkin (changes requested):
Since this is moby/sys repo which contain a bunch of packages, can you please prepend each commit title with `capability: ` prefix?

kolyshkin (approved):
LGTM

PR #165, opened 2024-09-30
capability: add separate API for ambient capabilities

lifubang:
In runtime-spec said[1]:
Runtimes SHOULD NOT fail if the container configuration requests capabilities that cannot be granted, for example, if the runtime operates in a restricted environment with a limited set of capabilities.
So this is the default mode when requesting capabilities. As a package used by `runc`, I think most of code about capabilities should be in this project, not in runc.

For the first edition of ambient implementation, though it masked the err because of a bug, but there was no `return` when got an error, it is different from the error handling in other place, so I think it was the author's original ideal to raise as many ambient cap sets as possible when got some errors.
There is no return in ambient caps apply:
https://github.com/moby/sys/blob/7b553f586f4d8789de3a26d20fb0beecb08b52bf/capability/capability/capability_linux.go#L477-L482
There is a return in other place:
https://github.com/moby/sys/blob/7b553f586f4d8789de3a26d20fb0beecb08b52bf/capability/capability/capability_linux.go#L451-L459

~~So, I think we should provide a choice to let users make the decision, for ambient cap sets apply, we should provide at least two modes:~~
~~1. Greedy mode: it should be the default mode to be compatible with the past, it means raise ambient caps as many as possible and return the last error;~~
~~2. Stop on error mode: it means that we should stop to raise other ambient caps when got an error.~~

So, I think we should provide the abilities to let users can make a decision to ignore the ambient cap raise/lower error or not.

[1] https://github.com/opencontainers/runtime-spec/blob/8f3fbc881602d85699e5c448634ec1288860d966/config.md?plain=1#L286-L292

### DIFF
--- a/capability/capability.go
+++ b/capability/capability.go
@@ -142,3 +142,18 @@ func NewFile2(path string) (Capabilities, error) {
 func LastCap() (Cap, error) {
 	return lastCap()
 }
+
+// AmbientRaise raises specified ambient capabilities for the calling process.
+func AmbientRaise(cap ...Cap) error {
+	return ambientRaise(cap...)
+}
+
+// AmbientLower lowers specified ambient capabilities for the calling process.
+func AmbientLower(cap ...Cap) error {
+	return ambientLower(cap...)
+}
+
+// AmbientClearAll lowers all ambient capabilities for the calling process.
+func AmbientClearAll() error {
+	return ambientClearAll()
+}
--- a/capability/capability_linux.go
+++ b/capability/capability_linux.go
@@ -117,6 +117,13 @@ func newPid(pid int) (c Capabilities, retErr error) {
 	return
 }
 
+func ignoreEINVAL(err error) error {
+	if errors.Is(err, syscall.EINVAL) {
+		err = nil
+	}
+	return err
+}
+
 type capsV3 struct {
 	hdr     capHeader
 	data    [2]capData
@@ -327,7 +334,7 @@ func (c *capsV3) Load() (err error) {
 	return
 }
 
-func (c *capsV3) Apply(kind CapType) (err error) {
+func (c *capsV3) Apply(kind CapType) error {
 	last, err := LastCap()
 	if err != nil {
 		return err
@@ -336,21 +343,17 @@ func (c *capsV3) Apply(kind CapType) (err error) {
 		var data [2]capData
 		err = capget(&c.hdr, &data[0])
 		if err != nil {
-			return
+			return err
 		}
 		if (1<<uint(CAP_SETPCAP))&data[0].effective != 0 {
 			for i := Cap(0); i <= last; i++ {
 				if c.Get(BOUNDING, i) {
 					continue
 				}
-				err = prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0)
+				// Ignore EINVAL since the capability may not be supported in this system.
+				err = ignoreEINVAL(prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0))
 				if err != nil {
-					// Ignore EINVAL since the capability may not be supported in this system.
-					if err == syscall.EINVAL { //nolint:errorlint // Errors from syscall are bare.
-						err = nil
-						continue
-					}
-					return
+					return err
 				}
 			}
 		}
@@ -359,33 +362,29 @@ func (c *capsV3) Apply(kind CapType) (err error) {
 	if kind&CAPS == CAPS {
 		err = capset(&c.hdr, &c.data[0])
 		if err != nil {
-			return
+			return err
 		}
 	}
 
 	if kind&AMBS == AMBS {
-		err = prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0)
-		if err != nil && err != syscall.EINVAL { //nolint:errorlint // Errors from syscall are bare.
-			// Ignore EINVAL as not supported on kernels before 4.3
-			return
+		// Ignore EINVAL as not supported on kernels before 4.3
+		err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0))
+		if err != nil {
+			return err
 		}
 		for i := Cap(0); i <= last; i++ {
 			if !c.Get(AMBIENT, i) {
 				continue
 			}
-			err = prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_RAISE, uintptr(i), 0, 0)
+			// Ignore EINVAL as not supported on kernels before 4.3
+			err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_RAISE, uintptr(i), 0, 0))
 			if err != nil {
-				// Ignore EINVAL as not supported on kernels before 4.3
-				if err == syscall.EINVAL { //nolint:errorlint // Errors from syscall are bare.
-					err = nil
-					continue
-				}
-				return
+				return err
 			}
 		}
 	}
 
-	return
+	return nil
 }
 
 func newFile(path string) (c Capabilities, err error) {
--- a/capability/capability_linux.go
+++ b/capability/capability_linux.go
@@ -368,7 +368,7 @@ func (c *capsV3) Apply(kind CapType) error {
 
 	if kind&AMBS == AMBS {
 		// Ignore EINVAL as not supported on kernels before 4.3
-		err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0))
+		err = ignoreEINVAL(ambientClearAll())
 		if err != nil {
 			return err
 		}
@@ -377,7 +377,7 @@ func (c *capsV3) Apply(kind CapType) error {
 				continue
 			}
 			// Ignore EINVAL as not supported on kernels before 4.3
-			err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_RAISE, uintptr(i), 0, 0))
+			err = ignoreEINVAL(ambientRaise(i))
 			if err != nil {
 				return err
 			}
@@ -385,6 +385,28 @@ func (c *capsV3) Apply(kind CapType) error {
 	}
 
 	return nil
+}
+
+func setAmbient(op uintptr, cap ...Cap) error {
+	for _, val := range cap {
+		err := prctl(pr_CAP_AMBIENT, op, uintptr(val), 0, 0)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ambientRaise(cap ...Cap) error {
+	return setAmbient(pr_CAP_AMBIENT_RAISE, cap...)
+}
+
+func ambientLower(cap ...Cap) error {
+	return setAmbient(pr_CAP_AMBIENT_LOWER, cap...)
+}
+
+func ambientClearAll() error {
+	return prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0, 0, 0)
 }
 
 func newFile(path string) (c Capabilities, err error) {
--- a/capability/capability_test.go
+++ b/capability/capability_test.go
@@ -137,7 +137,7 @@ func TestNewPid2Load(t *testing.T) {
 	}
 }
 
-func TestAmbientCapSet(t *testing.T) {
+func TestApplyCaps(t *testing.T) {
 	if runtime.GOOS != "linux" {
 		return
 	}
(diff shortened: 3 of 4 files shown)


lifubang:
Once #164 merged, I'll rebase this PR.

kolyshkin:
I think we should just make "greedy" mode a default.

Out of curiosity, have you checked what other implementations do? libcap C and Go versions, for example?

lifubang:
> Out of curiosity, have you checked what other implementations do? libcap C and Go versions, for example?

It seems like there is no function provided in `libcap C version` to raise more than one ambient caps in one time, so the users have to call `cap_set_ambient` one by one; Users have a choice in their code to ignore the error or not;

For go version, it provides such type function, there are two difference comparing with us:
1. It doesn't ignore any errors from syscall, includes `EINVAL`;
2. Like the name `Apply` on our project, it provides a single function `SetAmbient` to do this, but users can pass one or more ambient caps in params.
Though it returns error immediately, but I think if users want to ignore the SetAmbient errors, users have a choice to call it one by one.

The go version implementation code is:
```go
//go:uintptrescapes
func (sc *syscaller) setAmbient(enable bool, val ...Value) error {
	dir := uintptr(prCapAmbientLower)
	if enable {
		dir = prCapAmbientRaise
	}
	for _, v := range val {
		_, err := sc.prctlwcall6(prCapAmbient, dir, uintptr(v), 0, 0, 0)
		if err != nil {
			return err
		}
	}
	return nil
}
```

So I think we can have some choices:
1. Add `greedy` and `stop on error` mode like this PR;
2. Add two functions for ambient:
* func SetAmbient(enable bool, val ...Value) error
* func ResetAmbient() error
WDYT @kolyshkin

lifubang:
> So I think we can have some choices:
> 
> 1. Add `greedy` and `stop on error` mode like this PR;
> 2. Add two functions for ambient:
> 
> * func SetAmbient(enable bool, val ...Value) error
> * func ResetAmbient() error
>   WDYT @kolyshkin

In fact I’m prefer to the first solution, this is our own distinguishing feature, and I think it’s more convenient for the callers.

kolyshkin:
I'd like to follow the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) here. In other words, a separate function that changes a way errors are handled is a bit too much.

kolyshkin:
In fact, if we provide a function to apply capabilities one by one, a user can control what do to about errors.

lifubang:
> In fact, if we provide a function to apply capabilities one by one, a user can control what do to about errors.

Quiet agree, I add 3 APIs for ambient cap, maybe runc can use these APIs to support ambient cap set.
PTAL

capability/capability_linux.go
```diff
@@ -117,6 +117,11 @@ func newPid(pid int) (c Capabilities, retErr error) {
 	return
 }
 
+func ignorableError(err error) bool {
+	// Ignore EINVAL as not supported on kernels before 4.3
```
kolyshkin, line 121: See, the original code has two different comments about why we're ignoring EINVAL:
1. For `PR_CAPBSET_DROP`, it says "Ignore EINVAL since the capability may not be supported in this system."
2. For `PR_CAPSET_AMBIENT`, it says "Ignore EINVAL as not supported on kernels before 4.3."

With this change, you remove the first comment, moved the second one into this function, and thus it is now applicable to both cases, which is incorrect.

capability/capability_linux.go
```diff
@@ -336,21 +341,16 @@ func (c *capsV3) Apply(kind CapType) (err error) {
...
 				}
 				err = prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0)
-				if err != nil {
-					// Ignore EINVAL since the capability may not be supported in this system.
-					if err == syscall.EINVAL { //nolint:errorlint // Errors from syscall are bare.
-						err = nil
-						continue
-					}
-					return
+				if err != nil && !ignorableError(err) {
```
kolyshkin, line 352: Also, with this change, in case `kind` only has `BOUNDS` and the kernel returns `EINVAL` for the last PR_CAPBSET_DROP, it will be returned, while the code before your changes returns `nil` in the same case.

kolyshkin (changes requested):
Regarding the first commit -- the implementation in this PR is incorrect for a few reasons (see comments above). See https://github.com/moby/sys/pull/166 for a more correct implementation of what you're trying to achieve.

capability/capability_linux.go
lifubang: I don't think so. `ignorableError` is not the main reason of this refactor, removing the named return param is the main refactor, `err` is now a local val, not the return name. It will help to avoid causing other issues like the original issue of `ambient` implementation.

lifubang:
> Regarding the first commit -- the implementation in this PR is incorrect for a few reasons (see comments above). See #166 for a more correct implementation of what you're trying to achieve.

Fixed the comment.
Without this incorrect comment, do you think this implementation is valuable for runc?

kolyshkin:
> do you think this implementation is valuable for runc?

Yes, I think, we should set ambient capabilities separately and one by one, so we have a chance to ignore or warn errors.

I'd rename functions to `AmbientRaise`, `AmbientLower` and `AmbientLowerAll` or something like this.

lifubang:
> See #166 for a more correct implementation of what you're trying to achieve.

Changed like yours.

lifubang:
> I'd rename functions to `AmbientRaise`, `AmbientLower` and `AmbientLowerAll` or something like this.

Changed all function names.

capability/capability_linux.go
```diff
@@ -117,6 +117,14 @@ func newPid(pid int) (c Capabilities, retErr error) {
 	return
 }
 
+func ignoreEINVAL(err error) error {
+	// Ignore EINVAL since the op or the capability may not be supported in this system.
```
kolyshkin, line 121: Can you please remove this comment? Instead, explain why we're ignoring EINVAL when we call this function.

capability/capability_linux.go
```diff
@@ -336,21 +344,16 @@ func (c *capsV3) Apply(kind CapType) (err error) {
...
 		}
 		if (1<<uint(CAP_SETPCAP))&data[0].effective != 0 {
 			for i := Cap(0); i <= last; i++ {
 				if c.Get(BOUNDING, i) {
 					continue
 				}
-				err = prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0)
+				err = ignoreEINVAL(prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0))
 				if err != nil {
-					// Ignore EINVAL since the capability may not be supported in this system.
```
kolyshkin, line 348: Can we keep this comment? (move to to the line before one where `ignoreEINVAL` is called)?

capability/capability_linux.go
```diff
@@ -369,19 +372,14 @@ func (c *capsV3) Apply(kind CapType) (err error) {
 			if c.Get(AMBIENT, i) {
 				action = pr_CAP_AMBIENT_RAISE
 			}
-			err = prctl(pr_CAP_AMBIENT, action, uintptr(i), 0, 0)
+			err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, action, uintptr(i), 0, 0))
 			if err != nil {
-				// Ignore EINVAL as not supported on kernels before 4.3
```
kolyshkin, line 374: Same here, can we keep this comment? (move to to the line before one where `ignoreEINVAL` is called)?

kolyshkin:
I just realized this package API is flawed. You start with `NewPid2(pid)`, but then some capabilities can only be applied to current process (when `pid == 0` or pid == gettid()`). Meaning, if you do something like this:

```go
c, err := capability.NewPid(somePid)
...
c.Set(capability.AMBIENT, capability.CAP_CHOWN)
c.Apply(capability.AMBIENT)
```

The ambient capability will be applied to the current process, rather than the process identified by `somePid`.

😓 

Will open a new issue.

This PR is related, since it adds new methods to the above API. Since we can only raise/lower/clear own capabilities, Those should probably just be functions.

I'm also thinking we should only accept only one single capability as the argument, since otherwise we have the same problem with the errors (whether to continue or return an error if we caught one).

```go
func AmbientRaise(Cap) error
func AmbientLower(Cap) error
func AmbientClearAll() error
```

lifubang:
> The ambient capability will be applied to the current process, rather than the process identified by `somePid`.

:+1:  This is really a long time bug.


> This PR is related, since it adds new methods to the above API. Since we can only raise/lower/clear own capabilities, Those should probably just be functions.

Yes, for other users, to have a compatibility, we may also need to add a pid check in `Apply`.


> I'm also thinking we should only accept only one single capability as the argument, since otherwise we have the same problem with the errors (whether to continue or return an error if we caught one).

It's not very important, we can keep support a cap array, someone who want to ignore the error, he can pass a single cap one by one.
So, I'd prefer to support multi-caps param.

lifubang:
Fix the lint errors now, PTAL.

kolyshkin:
Can we have a fix for #165 as a separate PR? This one is becoming too big to consume.

capability/capability.go
```diff
@@ -142,3 +142,18 @@ func NewFile2(path string) (Capabilities, error) {
 func LastCap() (Cap, error) {
 	return lastCap()
 }
+
+// AmbientRaise is to raise caps for our own process
```
kolyshkin, line 146:
```suggestion
// AmbientRaise raises specified ambient capabilities for the calling process.
```

capability/capability.go
```diff
@@ -142,3 +142,18 @@ func NewFile2(path string) (Capabilities, error) {
...
+func AmbientRaise(cap ...Cap) error {
+	return ambientRaise(cap...)
+}
+
+// AmbientRaise is to lower caps for our own process
+func AmbientLower(cap ...Cap) error {
+	return ambientLower(cap...)
+}
+
+// AmbientRaise is to lower all caps for our own process
```
kolyshkin, line 156:
```suggestion
// AmbientClearAll lowers all ambient capabilities for the calling process.
```

capability/capability_linux.go
```diff
@@ -383,6 +387,32 @@ func (c *capsV3) Apply(kind CapType) error {
 	return nil
 }
 
+func setAmbient(raise bool, cap ...Cap) error {
```
kolyshkin, line 396: nit: instead of `raise bool` you can use `action uintptr` and then call it like this:
```go
func ambientRaise(cap ...Cap) error {
	return setAmbient(pr_CAP_AMBIENT_RAISE, cap...)
}

func ambientLower(cap ...Cap) error {
	return setAmbient(pr_CAP_AMBIENT_LOWER, cap...)
}
```

This removes an if and makes the code more straightforward/direct.

capability/capability_test.go
```diff
@@ -21,6 +21,16 @@ const (
 	maxLastCap = CAP_CHECKPOINT_RESTORE
 )
 
+func requirePCapSet(t *testing.T) {
```
kolyshkin, line 24: nit: can you use `t testing.TB` here instead (in case we'll add a benchmark later, this function can also be used from it).

capability/capability_test.go
```diff
@@ -21,6 +21,16 @@ const (
...
+	pid, err := NewPid(0)
```
kolyshkin, line 25: nit: use `NewPid2` and `Load` instead. `NewPid` is deprecated and we should not use it even in our own code, otherwise we're giving a bad example.

capability/capability_test.go
```diff
@@ -68,3 +78,56 @@ func TestListSupported(t *testing.T) {
...
+	}
+	requirePCapSet(t)
+
+	capBounding := []Cap{CAP_KILL, CAP_CHOWN, CAP_SYSLOG}
+	capPermitted := []Cap{CAP_KILL, CAP_CHOWN}
+	capEffective := []Cap{CAP_KILL}
+	capInheritable := []Cap{CAP_KILL, CAP_CHOWN}
+	capAmbient := []Cap{CAP_KILL, CAP_CHOWN}
+
+	pid, err := NewPid(0)
```
kolyshkin, line 94: ditto

capability/capability_test.go
```diff
@@ -68,3 +78,56 @@ func TestListSupported(t *testing.T) {
 		t.Fatalf("result is too short (got %d, want %d): +%v", len(list), minLen, list)
 	}
 }
+
+func TestAmbientCapSet(t *testing.T) {
```
kolyshkin, line 82: Hmm. You've just added `AmbientRaise`, `AmbientLower` and `AmbientClearAll` in the previous commit, but here you don't test those.

kolyshkin:
Needs a rebase (#164 is merged).

lifubang:
> Can we have a fix for #165 as a separate PR? This one is becoming too big to consume.

OK, I'll split this PR to two.

capability/capability_test.go
lifubang: Added one test now.

capability/capability_linux.go
```diff
@@ -387,6 +387,28 @@ func (c *capsV3) Apply(kind CapType) error {
...
+	}
+	return nil
+}
+
+func ambientRaise(cap ...Cap) error {
+	return setAmbient(pr_CAP_AMBIENT_RAISE, cap...)
+}
+
+func ambientLower(cap ...Cap) error {
+	return setAmbient(pr_CAP_AMBIENT_RAISE, cap...)
```
kolyshkin, line 405:
```suggestion
	return setAmbient(pr_CAP_AMBIENT_LOWER, cap...)
```

capability/capability_test.go
```diff
@@ -151,3 +151,63 @@ func TestAmbientCapSet(t *testing.T) {
 		}
 	}
 }
+
+func TestModifyAmbientCaps(t *testing.T) {
```
kolyshkin, line 155: You need to use testInChild for this one (see #173 and TestAmbientCapSet).

kolyshkin (changes requested):

kolyshkin:
Looks like this actually works now!
```
=== RUN   TestAmbientCapAPI
    capability_test.go:205: output from child:
        === RUN   TestAmbientCapAPI
--- PASS: TestAmbientCapAPI (0.00s)
```

kolyshkin (approved):
LGTM, thanks!

kolyshkin:
@lifubang thanks, only one nit: the last commit subject should be prefixed with `capability:`. Or, you can squash the last two commits since they are about the same thing.

kolyshkin:
> I'm thinking about the last form because I'm still not very convinced we should provide a function which sets multiple capabilities. It adds some complexity but on the caller's side it's just a simple `for` loop.

Also because we can provide `AmbientIsSet(which Cap) bool`, which can only check a single capability.

PR #166, opened 2024-10-06
capability: add/use ignoreEINVAL

kolyshkin:
This is mostly to simplify the code.


kolyshkin:
This is now part of #165.

PR #170, opened 2024-10-10
capability: test improvements

kolyshkin:
See commits for details.

@lifubang PTAL

### DIFF
--- a/capability/capability_test.go
+++ b/capability/capability_test.go
@@ -112,49 +112,42 @@ func TestAmbientCapSet(t *testing.T) {
 	}
 	requirePCapSet(t)
 
-	capBounding := []Cap{CAP_KILL, CAP_CHOWN, CAP_SYSLOG}
-	capPermitted := []Cap{CAP_KILL, CAP_CHOWN}
-	capEffective := []Cap{CAP_KILL}
-	capInheritable := []Cap{CAP_KILL, CAP_CHOWN}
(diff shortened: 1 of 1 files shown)


capability/capability_test.go
```diff
@@ -111,49 +112,42 @@ func TestAmbientCapSet(t *testing.T) {
...
 		t.Fatal(err)
 	}
-	pid.Set(BOUNDING, capBounding...)
-	pid.Set(PERMITTED, capPermitted...)
-	pid.Set(EFFECTIVE, capEffective...)
-	pid.Set(INHERITABLE, capInheritable...)
-	pid.Set(AMBIENT, capAmbient...)
-	if err = pid.Apply(CAPS | BOUNDING | AMBIENT); err != nil {
+
+	list := []Cap{CAP_KILL, CAP_CHOWN, CAP_SYS_CHROOT}
```
lifubang, line 120: From docker to kubernetes, they always fill each type cap sets with the same cap bits, I think maybe it's wrong. So I want to test cap sets with different caps, this is my thought.
But anyway, it's not worth to be different in this test.

lifubang (approved):

tianon (approved):

capability/capability_test.go
kolyshkin: Yes, I understand that (what I don't understand is why they do it that way -- for example, why drop bounding capabilities at all; but that is way out of scope for this PR).

PR #171, opened 2024-10-11
capability: can't raise ambient and drop bounding caps for other process

### LINKED ISSUE #168: [bug] capability: some errors related to pid
As capability package can be used to set other process's cap, but unfortunately except these two type caps:
1. Bounding cap: we can only drop capabilities from the bounding set via prctl(PR_CAPBSET_DROP), but this operation is only used for current process;
2. Ambient cap: we can only modify ambient cap set via prctl(PR_CAP_AMBIENT), but this operation is also only used for current process.

So if we use capability package for other process via `NewPid(pid>0)`, when we want to use `Apply` to modify target process's bounding and ambient cap set, it will modify current process's bouding and ambient cap set, rather than the target process.

The original post is here: https://github.com/moby/sys/pull/165#issuecomment-2400885314

lifubang:
There may be two ways to fix #168:
 1) Return an error when pid is not 0;
 2) Do raise and drop operation when pid is 0.

I think we should choose the second way, because besides the bug fix,
we can also have a compatibility with before.

### DIFF
--- a/capability/capability.go
+++ b/capability/capability.go
@@ -8,6 +8,8 @@
 // Package capability provides utilities for manipulating POSIX capabilities.
 package capability
 
+import "errors"
+
 type Capabilities interface {
 	// Get check whether a capability present in the given
 	// capabilities set. The 'which' value should be one of EFFECTIVE,
@@ -60,6 +62,11 @@ type Capabilities interface {
 	// effect.
 	Apply(kind CapType) error
 }
+
+var (
+	errBoundingNotMine = errors.New("not support drop bounding cap of other process")
+	errAmbientNotMine  = errors.New("not support modify ambient cap of other process")
+)
 
 // NewPid initializes a new [Capabilities] object for given pid when
 // it is nonzero, or for the current process if pid is 0.
--- a/capability/capability_test.go
+++ b/capability/capability_test.go
@@ -2,13 +2,13 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package capability_test
+package capability
 
 import (
+	"errors"
+	"os/exec"
 	"runtime"
 	"testing"
-
-	. "github.com/moby/sys/capability"
 )
 
 // Based on the fact Go 1.18+ supports Linux >= 2.6.32, and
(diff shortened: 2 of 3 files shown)


kolyshkin:
> There may be two ways to fix #168:
> 
> 1. Return an error when pid is not 0;
> 2. Do raise and drop operation when pid is 0.
> 
> I think we should choose the second way, because besides the bug fix, we can also have a compatibility with before.

There is a compatibility issue either way. With the second way, the behavior of this package has changed:
 - it used to ignore the `pid` argument and set capabilities for the current process;
 - now it doesn't do that.
 
 So, for some code which does, for example, `NewPid(os.Getpid())`, this will be a breaking change. What's worse, they won't know about it, as no error is returned.

So I think the first way is better. If someone is doing something wrong, it's better to return an error so they know they are doing it wrong, rather than silently stop doing something.

lifubang:
> So I think the first way is better. If someone is doing something wrong, it's better to return an error so they know they are doing it wrong, rather than silently stop doing something.

Agree, I have changed to return an error.

capability/capability.go
```diff
@@ -61,6 +63,11 @@ type Capabilities interface {
 	Apply(kind CapType) error
 }
 
+var (
+	ErrBoundingNotMine = errors.New("not support drop bounding cap of other process")
+	ErrAmbientNotMine  = errors.New("not support modify ambient cap of other process")
```
kolyshkin, line 68: Do we need these errors as part of a public API? If yes, why?

lifubang: Maybe someone use it to check the error type, like what to do in the test?
But maybe no, so I'll change it.

kolyshkin: > Maybe someone use it to check the error type, like what to do in the test?

For the test case, you can add `export_test.go` file having something like this:
```go
package capability

// Make some internal details available for tests.

var ErrBoundingNotMine = errBoundingNotMine
```

capability/capability_test.go
```diff
@@ -151,3 +150,35 @@ func TestAmbientCapSet(t *testing.T) {
...
+func TestApplyBoundingCapsForOtherProcess(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		return
+	}
+	requirePCapSet(t)
+
+	pid, err := NewPid(1)
+	if err != nil {
+		t.Fatal(err)
+	}
```
kolyshkin, line 179: Perhaps you can remove this part (i.e. make one test out of two).

capability/capability_test.go
```diff
@@ -151,3 +151,40 @@ func TestAmbientCapSet(t *testing.T) {
...
 	}
 }
+
+func TestApplyCapsForOtherProcess(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		return
+	}
+	requirePCapSet(t)
+
+	cmd := exec.Command("sleep", "sleep", "infinity")
```
kolyshkin, line 161: I think this should be 
```suggestion
	cmd := exec.Command("sleep", "infinity")
```

kolyshkin (changes requested):
We have a big problem

```
=== RUN   TestAmbientCapSet
--- PASS: TestAmbientCapSet (0.00s)
=== RUN   TestApplyCapsForOtherProcess
    capability_test.go:159: The test needs `CAP_SETPCAP`.
--- SKIP: TestApplyCapsForOtherProcess (0.00s)
```

TestAmbientCapSet removes some capabilities from the binary, and thus all other tests that require `CAP_SETPCAP` (or root) won't work.

Glad we found it now before it's too late.

The solution here is to spawn such test cases in a separate binary, so it won't affect others.

I will come up with a patch.

capability/capability_test.go
kolyshkin: The error went unnoticed because this test is not being run, because the previous test (TestAmbientCapSet) removes CAP_SET_PCAP from the test process. Oh well

kolyshkin: See [here](https://github.com/moby/sys/pull/171#pullrequestreview-2376989867)

kolyshkin:
> I will come up with a patch.

See #173.

kolyshkin:
> > I will come up with a patch.
> 
> See #173.

@lifubang can you help me and review #173?

kolyshkin:
Closing in favor of #174. @lifubang please help reviewing that one.

PR #172, opened 2024-10-18
Makefile: fix error from test

kolyshkin:
When running make test, there appears an error:

	/bin/sh: 1: test: go test -exec sudo: unexpected operator

or

	/bin/sh: line 1: test: go test -exec sudo: binary operator expected

(depending on which shell is used).

This is caused by CMD value containing extra quotes (from RUN_VIA_SUDO=-exec "sudo -n") which test sees as an extra option.

Enclose test argument in single quotes to fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ clean:
 
 .PHONY: foreach
 foreach: ## Run $(CMD) for every package.
-	@if test -z "$(CMD)"; then \
+	@if test -z '$(CMD)'; then \
 		echo 'Usage: make foreach CMD="commands to run for every package"'; \
 		exit 1; \
 	fi


kolyshkin:
@tianon I think you'll like this one :)

kolyshkin:
@thaJeztah PTAL

thaJeztah (approved):
DOH! That's a fun one indeed.

Hoping ... one day, `go test` to have a `--recursive` option to run all modules in the path recursively. This is such a pain all.

LGTM

tianon:
> @tianon I think you'll like this one :)

You're 100% correct, I _do_ like this one! :grin: :heart:

PR #174, opened 2024-10-22
capability: Apply: deny for another process

kolyshkin:
Current version of `capsV3.Apply` has a major problem: if you do something like this:
```go
	c, err := capability.NewPid(pid)
	...
	c.Set(capability.AMBIENT, capability.CAP_CHOWN)
	c.Apply(capability.AMBIENT)
```
then the ambient capability will be applied to the current process, rather than the process identified by pid. Same issue for `BOUNDS`.

For `CAPS` the situation is slightly different: `capset(2)` man page says:

> EPERM	The caller attempted to use capset() to modify the capabilities
> 	of a thread other than itself, but lacked sufficient privilege.
> 	For kernels supporting VFS capabilities, this is never permitted.

Here _kernels supporting VFS capabilities_ means most kernels >= v2.6.24, and all kernels >= v2.6.33. Since Go 1.18+ only supports Linux >= v2.6.32, this pretty much means "all kernels".

Meaning, `Apply(CAPS)` with non-zero pid will try `capset(2)` and return `EPERM`.

Let's return an error early if pid is set in Apply, and add a test case.

Fixes: #168
Closes: #171

### DIFF
--- a/capability/capability_linux.go
+++ b/capability/capability_linux.go
@@ -328,6 +328,9 @@ func (c *capsV3) Load() (err error) {
 }
 
 func (c *capsV3) Apply(kind CapType) (err error) {
+	if c.hdr.pid != 0 {
+		return errors.New("unable to modify capabilities of another process")
+	}
 	last, err := LastCap()
 	if err != nil {
 		return err
--- a/capability/capability_test.go
+++ b/capability/capability_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strings"
 	"testing"
 
 	. "github.com/moby/sys/capability"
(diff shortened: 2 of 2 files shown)


lifubang:
> Meaning, `Apply(CAPS)` with non-zero pid will try `capset(2)` and return `EPERM`.

But they also said:
That is, on
       kernels that have VFS capabilities support, when calling
       capset(), the only permitted values for hdrp->pid are 0 or,
       equivalently, the value returned by [gettid(2)](https://man7.org/linux/man-pages/man2/gettid.2.html).



> but lacked sufficient privilege.

It means this is permitted If we have a sufficient privilege?

kolyshkin:
> > Meaning, `Apply(CAPS)` with non-zero pid will try `capset(2)` and return `EPERM`.
> 
> But they also said: That is, on kernels that have VFS capabilities support, when calling capset(), the only permitted values for hdrp->pid are 0 or, equivalently, the value returned by [gettid(2)](https://man7.org/linux/man-pages/man2/gettid.2.html).

Theoretically, we should also allow settings CAPS when `pid == os.Gettid()`. Practically, since using a TID has the exactly same result as using `0` (in other words, there is no additional functionality when one uses a TID), it is easier for the library to just require `0` here and not handle the TID special case.

> > but lacked sufficient privilege.
> 
> It means this is permitted If we have a sufficient privilege?

It says _"For kernels supporting VFS capabilities, this is never permitted"_. This means that we never have a sufficient privilege. You can read the whole man page at e.g. https://man7.org/linux/man-pages/man2/capset.2.html (the other good source of information is https://man7.org/linux/man-pages/man7/capabilities.7.html).

kolyshkin:
@lifubang thank you for review, let me know what you think.

kolyshkin:
Merging this as we need to move forward

PR #176, opened 2024-10-31
capability: add separate ambient and bound API

kolyshkin:
The API being added is the same as in [kernel.org/pub/linux/libs/security/libcap/cap](https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap)
package, although the implementation is a bit simpler (here we only set
capabilities for the calling thread).

With this, we can switch runc to moby/sys/capability.

Co-authored-by: @lifubang 
Closes: #165

### DIFF
--- a/capability/capability_linux.go
+++ b/capability/capability_linux.go
@@ -371,7 +371,7 @@ func (c *capsV3) Apply(kind CapType) error {
 
 	if kind&AMBS == AMBS {
 		// Ignore EINVAL as not supported on kernels before 4.3
-		err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0))
+		err = ignoreEINVAL(resetAmbient())
 		if err != nil {
 			return err
 		}
@@ -380,7 +380,7 @@ func (c *capsV3) Apply(kind CapType) error {
 				continue
 			}
 			// Ignore EINVAL as not supported on kernels before 4.3
-			err = ignoreEINVAL(prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_RAISE, uintptr(i)))
+			err = ignoreEINVAL(setAmbient(true, i))
 			if err != nil {
 				return err
 			}
@@ -388,6 +388,32 @@ func (c *capsV3) Apply(kind CapType) error {
 	}
 
 	return nil
+}
+
+func getAmbient(c Cap) (bool, error) {
+	res, err := prctlRetInt(pr_CAP_AMBIENT, pr_CAP_AMBIENT_IS_SET, uintptr(c))
+	if err != nil {
+		return false, err
+	}
+	return res > 0, nil
+}
+
+func setAmbient(raise bool, caps ...Cap) error {
+	op := pr_CAP_AMBIENT_RAISE
+	if !raise {
+		op = pr_CAP_AMBIENT_LOWER
+	}
+	for _, val := range caps {
+		err := prctl(pr_CAP_AMBIENT, op, uintptr(val))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func resetAmbient() error {
+	return prctl(pr_CAP_AMBIENT, pr_CAP_AMBIENT_CLEAR_ALL, 0)
 }
 
 func newFile(path string) (c Capabilities, err error) {
--- a/capability/capability_linux.go
+++ b/capability/capability_linux.go
@@ -354,7 +354,7 @@ func (c *capsV3) Apply(kind CapType) error {
 					continue
 				}
 				// Ignore EINVAL since the capability may not be supported in this system.
-				err = ignoreEINVAL(prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0))
+				err = ignoreEINVAL(dropBound(i))
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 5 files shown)


kolyshkin:
@lifubang PTAL (this is similar to #165 plus
 - adds `GetAmbient` (for completeness);
 - makes the API similar to that of [cap](https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap) package;
 - some nits here and there.

capability/capability_linux.go
```diff
@@ -117,6 +117,13 @@ func newPid(pid int) (c Capabilities, retErr error) {
 	return
 }
 
+func ignoreEINVAL(err error) error {
+	if errors.Is(err, syscall.EINVAL) {
+		err = nil
```
thaJeztah, line 122: Reminds me that I stumbled on this package which had a "consider integrating with moby/sys todo; https://github.com/moby/moby/blob/dc225798cbddebd47bfaa0fd8337d145c91fc6ba/internal/unix_noeintr/fs_unix.go#L3-L5

capability/capability_test.go
```diff
@@ -227,3 +228,88 @@ func TestApplyOtherProcess(t *testing.T) {
...
+	for _, cap := range list {
+		want := false
+		got, err := GetAmbient(cap)
+		if err != nil {
+			log.Fatalf("GetAmbient(%s): want nil, got error %v", cap, err)
+		} else if want != got {
+			log.Fatalf("Get(AMBIENT, %s): want %v, got %v", cap, want, got)
+		}
+	}
+	os.Exit(0)
```
thaJeztah, line 314: LOL, had to look twice, but then saw it's called from `testInChild`

thaJeztah (approved):
LGTM

thaJeztah:
@AkihiroSuda @tianon ptal

capability/syscall_linux.go
```diff
@@ -48,8 +48,8 @@ const (
 	pr_CAP_AMBIENT_CLEAR_ALL = uintptr(4)
 )
 
-func prctl(option int, arg2, arg3, arg4, arg5 uintptr) (err error) {
-	_, _, e1 := syscall.Syscall6(syscall.SYS_PRCTL, uintptr(option), arg2, arg3, arg4, arg5, 0)
+func prctl(option int, arg2, arg3 uintptr) (err error) {
+	_, _, e1 := syscall.Syscall(syscall.SYS_PRCTL, uintptr(option), arg2, arg3)
```
thaJeztah: I could've sworn these numbered ones were deprecated, but that was for _windows_ there they added a `SyscallN()`; https://github.com/golang/go/blob/6d39245514c675cdea5c7fd7f778e97bf0728dd5/src/syscall/dll_windows.go#L27-L45

capability/capability_linux.go
kolyshkin: That thing in there is something entirely different -- it's a retry-on-EINTR, and here we have ignore-EINVAL.

As to autogenerating stuff, even github.com/golang/go doesn't do it -- they use helper functions here and there but it's all manual.

capability/capability_test.go
kolyshkin: Yeah, it's somewhat complicated here since we modify capabilities and if we do it in a current process the following tests are toasted.

capability/syscall_linux.go
kolyshkin: Wonder why they did this change only for Windows. Perhaps it's better aligned to native windows apis.

capability/capability_linux.go
```diff
@@ -339,21 +346,17 @@ func (c *capsV3) Apply(kind CapType) (err error) {
...
+			return err
 		}
 		if (1<<uint(CAP_SETPCAP))&data[0].effective != 0 {
 			for i := Cap(0); i <= last; i++ {
 				if c.Get(BOUNDING, i) {
 					continue
 				}
-				err = prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0, 0, 0)
+				// Ignore EINVAL since the capability may not be supported in this system.
+				err = ignoreEINVAL(prctl(syscall.PR_CAPBSET_DROP, uintptr(i), 0))
```
tianon, line 357: Not a huge deal (certainly not a dealbreaker), but it does seem weird to still use `prctl` directly here when all the rest have switched to your new wrappers -- was there a reason not to use `dropBound` here?

capability/capability_test.go
tianon: I'm way outside my understanding, but isn't that what `runtime.LockOSThread` is for/does?
(makes sure the capabilities are adjusted for the thread we're on, and then if we don't unlock the thread, it exits instead of being reused?)

tianon (approved):
Looks good (and IMO totally fine to merge as-is) -- I left a couple questions, but I don't think they're super important.

capability/capability_linux.go
kolyshkin: Nice catch, thanks!

I added GetBound/DropBound at the last moment, and this here is part of an earlier patch. Still makes sense to use dropBound here -- updated.

capability/capability_test.go
kolyshkin: We are actually re-executing the test binary here (or, rather, in `testInChild`), so here we have a separate process (not a bare go thread) running.

It never occurred to be we can test all this in a separate go thread (rather than a process). Let me give it a try :)

kolyshkin: I quickly tried that and it did't work. I'll add it to TODO and merge this as is for now, since it's just a test case.

kolyshkin: Ah, the `runtime.LockOSThread()`, as added, is needed so that Go runtime won't suddenly switch the goroutine to other OS thread while we run the test.

Practically, for users of this package, it means we need to change capabilities (in a locked OS thread) and then re-exec (which is what runc does).

tianon: Yeah, I was looking at capabilities being a per thread attribute and scratching my beard 😂  sorry it didn't work out, but at least now you know!

Issue #146, opened 2024-08-08
Porting uid/gid mapping ID conversion logic from idtools


henry118:
I am working on a feature in containerd to support multiple uid/gid mappings https://github.com/containerd/containerd/pull/10307. 

My current implementation of the [ID conversion logic](https://github.com/containerd/containerd/pull/10307/files#diff-e6f16a611e7db5ba3ba07ade868a2594d5ca3323f45c661aca8b1a4f929f8151) is based on the [idtools](https://github.com/moby/moby/blob/master/pkg/idtools/idtools.go) code in Moby.

Since the `moby/sys` package also contains code for ID mapping, such as parsing `/etc/{subuid,subuid}` and the [IDMap type](https://github.com/moby/sys/blob/main/user/user.go#L53-L57), would it be nice to also include the logic for ID conversions (ie. `ToHost()`/`ToContainer()` funcs)? This way containerd can have much cleaner implementation.

WDYT?

PR #182, opened 2025-01-10
Merge functionality from `github.com/moby/moby/pkg/idtools` into `user`

dmcgowan:
Merges the idtools package functionality from github.com/moby/moby/pkg/idtools into the user package. The package was previously trimmed down and is simplified further to better fit `moby/sys`.

Now the package interface will just expand with:
- `MkdirAllAndChown` with  `WithOnlyNew` as a `MkdirOpt`
- `MkdirAndChown` also supporting the opts
- `IdentityMapping` its functions `RootPair`, `ToHost`, `ToContainer`, `Empty`

Note the `Identity` type is gone since `SID` was not used in the package or by any of the functions. It is more common to just pass along the uid and gid.

### DIFF
--- a/user/idtools_unix_test.go
+++ b/user/idtools_unix_test.go
@@ -1,17 +1,15 @@
 //go:build !windows
 
-package idtools
+package user
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"golang.org/x/sys/unix"
-	"gotest.tools/v3/assert"
-	is "gotest.tools/v3/assert/cmp"
-	"gotest.tools/v3/skip"
 )
 
 type node struct {
@@ -20,12 +18,8 @@ type node struct {
 }
 
 func TestMkdirAllAndChown(t *testing.T) {
-	RequiresRoot(t)
-	dirName, err := os.MkdirTemp("", "mkdirall")
-	if err != nil {
-		t.Fatalf("Couldn't create temp dir: %v", err)
-	}
-	defer os.RemoveAll(dirName)
+	requiresRoot(t)
+	dirName := t.TempDir()
 
 	testTree := map[string]node{
 		"usr":              {0, 0},
@@ -40,7 +34,7 @@ func TestMkdirAllAndChown(t *testing.T) {
 	}
 
 	// test adding a directory to a pre-existing dir; only the new dir is owned by the uid/gid
-	if err := MkdirAllAndChown(filepath.Join(dirName, "usr", "share"), 0o755, Identity{UID: 99, GID: 99}); err != nil {
+	if err := MkdirAllAndChown(filepath.Join(dirName, "usr", "share"), 0o755, 99, 99); err != nil {
 		t.Fatal(err)
 	}
 	testTree["usr/share"] = node{99, 99}
@@ -48,12 +42,10 @@ func TestMkdirAllAndChown(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := compareTrees(testTree, verifyTree); err != nil {
-		t.Fatal(err)
-	}
+	compareTrees(t, testTree, verifyTree)
 
 	// test 2-deep new directories--both should be owned by the uid/gid pair
-	if err := MkdirAllAndChown(filepath.Join(dirName, "lib", "some", "other"), 0o755, Identity{UID: 101, GID: 101}); err != nil {
+	if err := MkdirAllAndChown(filepath.Join(dirName, "lib", "some", "other"), 0o755, 101, 101); err != nil {
 		t.Fatal(err)
 	}
 	testTree["lib/some"] = node{101, 101}
@@ -62,12 +54,10 @@ func TestMkdirAllAndChown(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := compareTrees(testTree, verifyTree); err != nil {
-		t.Fatal(err)
-	}
+	compareTrees(t, testTree, verifyTree)
 
 	// test a directory that already exists; should be chowned, but nothing else
-	if err := MkdirAllAndChown(filepath.Join(dirName, "usr"), 0o755, Identity{UID: 102, GID: 102}); err != nil {
+	if err := MkdirAllAndChown(filepath.Join(dirName, "usr"), 0o755, 102, 102); err != nil {
 		t.Fatal(err)
 	}
 	testTree["usr"] = node{102, 102}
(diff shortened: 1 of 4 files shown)


austinvazquez:
Resolves https://github.com/moby/sys/issues/146?

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+	gid int
+}
+
+func TestMkdirAllAndChown(t *testing.T) {
+	RequiresRoot(t)
+	dirName, err := os.MkdirTemp("", "mkdirall")
+	if err != nil {
+		t.Fatalf("Couldn't create temp dir: %v", err)
+	}
+	defer os.RemoveAll(dirName)
```
kolyshkin, line 26: nit: use `t.TempDir`

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+	compareTrees(t, testTree, verifyTree)
+}
+
+func TestMkdirAllAndChownNew(t *testing.T) {
+	RequiresRoot(t)
+	dirName, err := os.MkdirTemp("", "mkdirnew")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dirName)
```
kolyshkin, line 81: ditto (use `t.TempDir`)

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+
+			if err := MkdirAllAndChown(tc.in, 0o755, expectedUIDGID, expectedUIDGID, WithOnlyNew); err != nil {
+				t.Fatal(err)
+			}
+
+			for _, p := range tc.out {
+				s := &unix.Stat_t{}
+				if err := unix.Stat(p, s); err != nil {
+					t.Error(err)
```
kolyshkin, line 199: nit: unix errors are bare, so maybe `t.Errorf("stat %v: %v", p, err)`

(not really important here since this is a test)

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+}
+
+func TestMkdirAndChown(t *testing.T) {
+	RequiresRoot(t)
+	dirName, err := os.MkdirTemp("", "mkdir")
+	if err != nil {
+		t.Fatalf("Couldn't create temp dir: %v", err)
+	}
+	defer os.RemoveAll(dirName)
```
kolyshkin, line 237: ditto (use `t.TempDir`)

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+	testTree["usr"] = node{99, 99}
+	verifyTree, err := readTree(dirName, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	compareTrees(t, testTree, verifyTree)
+
+	// create a subdir under a dir which doesn't exist--should fail
+	if err := MkdirAndChown(filepath.Join(dirName, "usr", "bin", "subdir"), 0o755, 102, 102); err == nil {
+		t.Fatalf("Trying to create a directory with Mkdir where the parent doesn't exist should have failed")
```
kolyshkin, line 259: nit: `t.Fatal`

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+}
+
+func buildTree(base string, tree map[string]node) error {
+	for path, node := range tree {
+		fullPath := filepath.Join(base, path)
+		if err := os.MkdirAll(fullPath, 0o755); err != nil {
+			return fmt.Errorf("couldn't create path: %s; error: %w", fullPath, err)
+		}
+		if err := os.Chown(fullPath, node.uid, node.gid); err != nil {
+			return fmt.Errorf("couldn't chown path: %s; error: %w", fullPath, err)
```
kolyshkin, line 281: nit: no need to wrap `os` errors as they already have context (path and operation).

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+		t.Fatal(err)
+	}
+	compareTrees(t, testTree, verifyTree)
+}
+
+func buildTree(base string, tree map[string]node) error {
+	for path, node := range tree {
+		fullPath := filepath.Join(base, path)
+		if err := os.MkdirAll(fullPath, 0o755); err != nil {
+			return fmt.Errorf("couldn't create path: %s; error: %w", fullPath, err)
```
kolyshkin, line 278: it: no need to wrap os errors as they already have context (path and operation).

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return nil
+}
+
+func readTree(base, root string) (map[string]node, error) {
+	tree := make(map[string]node)
+
+	dirInfos, err := os.ReadDir(base)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't read directory entries for %q: %w", base, err)
```
kolyshkin, line 292: ditto (no need to wrap `os` errors)

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+// TestMkdirIsNotDir checks that mkdirAs() function (used by MkdirAll...)
+// returns a correct error in case a directory which it is about to create
+// already exists but is a file (rather than a directory).
+func TestMkdirIsNotDir(t *testing.T) {
+	file, err := os.CreateTemp("", t.Name())
+	if err != nil {
+		t.Fatalf("Couldn't create temp dir: %v", err)
+	}
+	defer os.Remove(file.Name())
```
kolyshkin, line 399: ditto (use `t.TempDir`)

user/idtools_unix_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	err = mkdirAs(file.Name(), 0o755, 0, 0, false, false)
+	if expected := "mkdir " + file.Name() + ": not a directory"; err.Error() != expected {
+		t.Fatalf("expected error: %v, got: %v", expected, err)
+	}
+}
+
+func RequiresRoot(t *testing.T) {
```
kolyshkin, line 407: nit: this should not be a public function.

kolyshkin (commented):
Left some nits (can be addressed as a followup).

My biggest question though is why not place it into a separate `moby/sys/idtools` package?

tianon:
This new code is unrelated to the existing `user` code, right? Can we please use a different package for it?

tianon:
See also https://github.com/moby/sys/pull/145

tianon:
I guess with no new deps it is a bit of a different case 🤔👍

dmcgowan:
> This new code is unrelated to the existing user code, right? Can we please use a different package for it?

Well that is what is up for discussion. I think it makes sense for this logic to go along with the `IDMap` type, however, it is arguable that the ID map logic itself doesn't fit next to the uid lookup and goes better with userns stuff or elsewhere.

thaJeztah (approved):
LGTM

PR #184, opened 2025-03-12
user: add special container user constants for windows to sys/user

jsternberg:


crazy-max:
relates to https://github.com/moby/buildkit/pull/5791#pullrequestreview-2650964109

kolyshkin:
It would be nice to at least add a comment telling something about where these are coming from (I know they are not documented etc. -- but at least provide some sources).

thaJeztah:
Let me leave a "request changes", because i don't think we'd be merging this currently.

As I commented on other PRs, I think these "magic" users should be considered a contract for Windows containers. And given that there's only 1 party that provides Windows base images, those should be considered a specification, and these users to be a part of that.

So what I would like to see;

- Microsoft to include these somewhere in their official documentation
- Given that they're defined in Microsoft/hcsshim; a set of properly documented consts (or similar) outlining these users (and their purpose)
- Ideally, the OCI specification to also include these (as, per the above, it's effectively a specification for Windows images)

I want to avoid Moby, or BuildKit to become the "source of truth" for these consts, because we don't own this specification, and we don't want others to depend on those, so preferably keep them either internal, or IF we export them, to be documented with a reference to the canonical source of truth / specification.

thaJeztah (changes requested):
see above

jsternberg:
We merged the version of this that has these constants in buildkit. I think @thaJeztah is correct that this should go somewhere else and buildkit is a good place to put it temporarily.

PR #185, opened 2025-04-01
integrate pkg/atomicwriter

thaJeztah:
- relates to https://github.com/moby/moby/pull/49608
- relates to https://github.com/moby/moby/issues/32989

### integrate pkg/atomicwriter

This integrates the atomicwriter package from moby at commit;
[d7b743b8569e253b61ea8d70e75146ff35d50417][1].

Migration was done using the following steps:

```bash
brew install git-filter-repo

cd ~/Projects
git clone https://github.com/moby/moby.git moby_temp
cd moby_temp

# commit taken from
git rev-parse --verify HEAD
d7b743b8569e253b61ea8d70e75146ff35d50417

git filter-repo --analyze

# remove all code, except for 'pkg/atomicwriter', and rename to /atomicwriter
git filter-repo \
  --path 'pkg/ioutils/fswriters.go' \
  --path 'pkg/ioutils/fswriters_test.go' \
  --path 'pkg/atomicwriter' \
  --path-rename pkg/atomicwriter:atomicwriter

git checkout -b integrate_atomicwriter

git remote add moby_temp ~/Projects/moby_temp
git fetch moby_temp

# merge the upstream code
git merge --allow-unrelated-histories --signoff -S moby_temp/master
```

[1]: https://github.com/moby/moby/commit/d7b743b8569e253b61ea8d70e75146ff35d50417

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES ?= capability mountinfo mount reexec sequential signal symlink user userns # IMPORTANT: when updating this list, also update the conditional one in .github/workflows/test.yml
+PACKAGES ?= atomicwriter capability mountinfo mount reexec sequential signal symlink user userns # IMPORTANT: when updating this list, also update the conditional one in .github/workflows/test.yml
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
 	freebsd/amd64 openbsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
@@ -29,9 +29,12 @@ test: test-local
 test: CMD=go test $(RUN_VIA_SUDO) -v -coverprofile=coverage.txt -covermode=atomic .
 test: foreach
 
-# Test the mount module against the local mountinfo source code instead of the
-# release specified in its go.mod. This allows catching regressions / breaking
-# changes in mountinfo.
+# Some modules in this repo have interdependencies:
+#  - mount depends on mountinfo
+#  - atomicwrite depends on sequential
+#
+# The code below tests these modules against their local dependencies
+# to catch regressions / breaking changes early.
 .PHONY: test-local
 test-local: MOD = -modfile=go-local.mod
 test-local:
@@ -39,6 +42,10 @@ test-local:
 	# Run go mod tidy to make sure mountinfo dependency versions are met.
 	cd mount && go mod tidy $(MOD) && go test $(MOD) $(RUN_VIA_SUDO) -v .
 	$(RM) mount/go-local.*
+	echo 'replace github.com/moby/sys/sequential => ../sequential' | cat atomicwriter/go.mod - > atomicwriter/go-local.mod
+	# Run go mod tidy to make sure sequential dependency versions are met.
+	cd atomicwriter && go mod tidy $(MOD) && go test $(MOD) $(RUN_VIA_SUDO) -v .
+	$(RM) atomicwriter/go-local.*
 
 .PHONY: lint
 lint: $(BINDIR)/golangci-lint
--- a/atomicwriter/atomicwriter.go
+++ b/atomicwriter/atomicwriter.go
@@ -1,0 +1,245 @@
+// Package atomicwriter provides utilities to perform atomic writes to a
+// file or set of files.
+package atomicwriter
+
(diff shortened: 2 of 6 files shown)


atomicwriter/atomicwriter.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func validateFileMode(mode os.FileMode) error {
+	switch {
+	case mode.IsRegular():
+		return nil // Regular file
+	case mode&os.ModeDir != 0:
+		return errors.New("cannot write to a directory")
+	// TODO(thaJeztah): decide whether we want to disallow symlinks or to follow them.
+	// case mode&os.ModeSymlink != 0:
```
cpuguy83, line 46: Maybe we can get rid of this todo and just error out unless we are explicitly going to add a case to follow symlinks along with a test.

atomicwriter/atomicwriter.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+		return errors.New("cannot write to a block device file")
+	case mode&os.ModeSetuid != 0:
+		return errors.New("cannot write to a setuid file")
+	case mode&os.ModeSetgid != 0:
+		return errors.New("cannot write to a setgid file")
+	case mode&os.ModeSticky != 0:
+		return errors.New("cannot write to a sticky bit file")
+	default:
+		// Unknown file mode; let's assume it works
```
cpuguy83, line 64: Seems gutsy.

cpuguy83 (changes requested):
I get this is just copy/paste from moby/moby, but probably good to deal with some of the unknown edge cases.

atomicwriter/atomicwriter.go
thaJeztah: It's a slightly tricky one; if we want this writer to be a drop-in for a regular `os.WriteFile`, then we should probably follow the symlink. We could resolve the symlink when creating the writer, then use that value as destination, but of course we may need to check where it's currently used (in case this would open up some symlink attack vector in moby?)

Erroring out would (in the meantime?) also be fine; currently it's kinda horrible, because if the destination is a symlink, the `os.Rename` just replaces the symlink with a file, which is not very expected I guess.

Makefile
```diff
@@ -29,7 +29,7 @@ test: test-local
 test: CMD=go test $(RUN_VIA_SUDO) -v -coverprofile=coverage.txt -covermode=atomic .
 test: foreach
 
-# Test the mount module against the local mountinfo source code instead of the
+# Test the mount module against the local sequential and mountinfo source code instead of the
```
kolyshkin, line 32: This is not entirely correct (mount is not tested against sequential).

I'd replace all this with something like:

```make
# Some modules in this repo have interdependencies:
#  - mount depends on mountinfo
#  - atomicwrite depedns on sequential
#
# The code below tests these modules against their local dependencies,
# to catch regressions / breaking changes early.
```

thaJeztah: Oh! Good one; yes will update.

FWIW; I opened a PR in moby to address Brian's comments; https://github.com/moby/moby/pull/49741

(I'm considering to explore if we could / should make the "write to symlink" work, but with it being changed to an error in that PR, it's something we can always to in a follow-up)

If those are accepted and merged, I'll re-do this PR. Let me move it back to draft until I'm ready to do so. 👍

thaJeztah:
Alright; updated; @cpuguy83 @kolyshkin @dmcgowan PTAL 🤗

Makefile
thaJeztah: Applied your suggestions 👍

atomicwriter/atomicwriter.go
thaJeztah: fixed in https://github.com/moby/moby/pull/49741

atomicwriter/atomicwriter.go
thaJeztah: fixed in https://github.com/moby/moby/pull/49741 - now producing an error (at least for now) for symlinked target

cpuguy83 (approved):

thaJeztah:
Thanks! I'll go ahead and bring this in; might tag a first version later.

kolyshkin (approved):
LGTM, thanks

PR #186, opened 2025-04-04
ci: add go1.24, add macOS 15, remove ubuntu 20.04 (EOL) and update golangci-lint to v2.0.2

thaJeztah:
- [x] depends on https://github.com/moby/sys/pull/187

PR #187, opened 2025-04-04
fix linting issues for golangci-lint v2

thaJeztah:
- relates to https://github.com/moby/sys/pull/186

We could suppress this linter, as using w.File.xxx makes it slightly more explicit, but probably fine as well.

```
Error: atomicwriter/atomicwriter.go:205:11: QF1008: could remove embedded field "File" from selector (staticcheck)
	err := w.File.Sync()
	         ^
```

```
Error: mountinfo/mounted_linux_test.go:282:6: QF1001: could apply De Morgan's law (staticcheck)
		if !(tc.isNotExist && errors.Is(err, os.ErrNotExist)) {
		   ^
Error: mountinfo/mounted_linux_test.go:353:8: QF1001: could apply De Morgan's law (staticcheck)
				if !(tc.isNotExist && errors.Is(err, os.ErrNotExist)) {
				   ^
```

```
Error: capability/syscall_linux.go:144:2: QF1003: could use tagged switch on data.version (staticcheck)
	if data.version == 1 {
	^
```

```
Error: user/user.go:447:16: ST1005: error strings should not be capitalized (staticcheck)
			return nil, fmt.Errorf("Unable to find additional groups %v: %w", additionalGroups, err)
			            ^
Error: user/user.go:471:17: ST1005: error strings should not be capitalized (staticcheck)
				return nil, fmt.Errorf("Unable to find group %s: %w", ag, ErrNoGroupEntries)
				            ^
```

### DIFF
--- a/capability/syscall_linux.go
+++ b/capability/syscall_linux.go
@@ -141,16 +141,17 @@ func setVfsCap(path string, data *vfscapData) (err error) {
 		return
 	}
 	var size uintptr
-	if data.version == 1 {
+	switch data.version {
+	case 1:
 		data.magic = vfsCapVer1
 		size = vfscapDataSizeV1
-	} else if data.version == 2 {
+	case 2:
 		data.magic = vfsCapVer2
 		if data.effective[0] != 0 || data.effective[1] != 0 {
 			data.magic |= vfsCapFlageffective
 		}
 		size = vfscapDataSizeV2
-	} else {
+	default:
 		return syscall.EINVAL
 	}
 	_, _, e1 := syscall.RawSyscall6(syscall.SYS_SETXATTR, uintptr(unsafe.Pointer(_p0)), uintptr(unsafe.Pointer(_vfsXattrName)), uintptr(unsafe.Pointer(data)), size, 0, 0)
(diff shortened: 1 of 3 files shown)


austinvazquez (approved):

thaJeztah:
Dropped the first commit; I'll bring this one in and rebase the other one.

kolyshkin (commented):
LGTM

Believe it or not, but this is almost[1] line-for-line identical my patchset which I prepared last Friday but did not open a PR :)

[1]: except the bit in https://github.com/moby/sys/pull/188

thaJeztah:
Ha! Yes, most of the changes seemed reasonable; I was _slightly_ on the fence on the "apply De Morgan's law" changes in https://github.com/moby/sys/pull/187/commits/8d553c09dc5e9ec44c52fc98ccccfeffb80450b4, but then again, either before and after would need a short comment (already in place) to outline the logic, so I took that one as "OK, let's make the linter happy".

I didn't like the "use embedded struct directly". At least; not in this specific case (there's situations where it's OK, but this one was not one of them).

PR #188, opened 2025-04-07
.golangci.yml: remove unused exceptions

kolyshkin:
This is a followup to #186.

This repo:
 - does not have any paths that are excluded;
 - does not have any generated code.

 Also, neither "common-false-positives" nor "legacy" presets suppress
 any warnings, so remove those as well.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,20 +1,15 @@
 version: "2"
+
 linters:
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM, thanks! 🙏

PR #189, opened 2025-04-08
Move functionality from `github.com/docker/docker/profiles/seccomp`

jsternberg:
Moves the `github.com/docker/docker/profiles/seccomp` package into
`github.com/moby/sys/seccomp`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,7 +26,7 @@ jobs:
       run: |
         # This corresponds with the list in Makefile:1, but omits the "userns"
         # and "capability" modules, which require go1.21 as minimum.
-        echo 'PACKAGES=atomicwriter mountinfo mount reexec sequential signal symlink user' >> $GITHUB_ENV
+        echo 'PACKAGES=atomicwriter mountinfo mount reexec seccomp sequential signal symlink user' >> $GITHUB_ENV
     - name: go mod tidy
       run: |
         make foreach CMD="go mod tidy"
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PACKAGES ?= atomicwriter capability mountinfo mount reexec sequential signal symlink user userns # IMPORTANT: when updating this list, also update the conditional one in .github/workflows/test.yml
+PACKAGES ?= atomicwriter capability mountinfo mount reexec seccomp sequential signal symlink user userns # IMPORTANT: when updating this list, also update the conditional one in .github/workflows/test.yml
 BINDIR ?= _build/bin
 CROSS ?= linux/arm linux/arm64 linux/ppc64le linux/s390x \
 	freebsd/amd64 openbsd/amd64 darwin/amd64 darwin/arm64 windows/amd64
--- a/seccomp/default.json
+++ b/seccomp/default.json
@@ -1,0 +1,833 @@
+{
+	"defaultAction": "SCMP_ACT_ERRNO",
+	"defaultErrnoRet": 1,
+	"archMap": [
+		{
+			"architecture": "SCMP_ARCH_X86_64",
+			"subArchitectures": [
+				"SCMP_ARCH_X86",
+				"SCMP_ARCH_X32"
+			]
+		},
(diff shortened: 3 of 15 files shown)


kolyshkin (commented):
Generally, here in moby/sys we have low-level stuff which do not have tons of external dependencies (some of them depend on others in this repo, some need x/sys/unix, and that's all). That alone makes me think this is not the best place.

Having said that, it might make sense to create kernelversion package here, for example.

If there's something that can be added to https://github.com/seccomp/libseccomp-golang (which I happen to be a maintainer of), I'm also open to suggestions.

thaJeztah:
Thanks!

w.r.t. the "kernel version parsing"; _possibly_? The reason it was made internal is that we originally had a kernel-parsing package in moby, but we didn't want it to be an all-purpose one (people started to contribute "macOS kernel version parsing" and the like). I also moved it internal into this package to try and keep it self-contained with the possibility of moving the seccomp and AppArmor profiles to a separate module.

(it looks like `gotest.tools` made its way back into the package for some assertions, which we definitely should look if strictly needed, and otherwise just use stdlib for this again)


For a bit of context; we are currently working on trying to fix the circular dependency between BuildKit and Moby; we're _almost_ there, but the seccomp profile was one of the last remaining ones. The reason BuildKit uses the seccomp profile from Moby is so that we make sure the profile is "in sync" with the one used by Moby itself; before that we used the profile from containerd (the containerd profile started as a fork of the one in Moby); while we TRY to keep those profiles in sync as well, it sometimes differed slightly with Moby (see https://github.com/moby/buildkit/pull/1807).

An "easy" temporary solution would be for BuildKit to switch back to the containerd profile; however, there may be cases where Moby has changes in the profile that have not (yet) made their way into the containerd release.

All that said; there's been various discussions related to the profile, because (as shown above), we try to keep them all in sync, and having multiple forks doesn't help there.

So my ideal would be

- for the profile(s) (seccomp, apparmor) to live in a separate module
- align some of the differences in implementation (I think the containerd one had a slightly nicer interface to use, which basically was "apply this to this OCI spec", but I'd have to look at that again)
- the _moby_ package also defines a file-format (JSON), which I'm not sure the containerd package provides; not strictly critical for the seccomp profile as a whole, but probably also OK to keep as part of the same module (but could be a package in it)

And with the above, both Moby, BuildKit, and containerd could use the same module for the seccomp profile; also allowing the profile to have its own versioning (something to discuss though if we want that based on the module version, or to have some "version" selector for applying the profile if there's possibly changes in the profile that could break some scenarios).

One of the things we _SHOULD_ look at is to finish that ever-lasting issue of the `EPERM` as default (I have an internal ticket related to that, and probably should put it in the Moby issue tracker);

- https://github.com/moby/moby/issues/42871
- possibly we could use the work that was done in `containers/common` (most notably https://github.com/containers/common/pull/573)

☝️ whatever we do on that matter, we should try to preserve git history for all of those changes, as there's various cases where history contains very relevant information (why was a `syscall` added / removed? what was the motivation and decision around it?)

kolyshkin:
> One of the things we _SHOULD_ look at is to finish that ever-lasting issue of the `EPERM` as default (I have an internal ticket related to that, and probably should put it in the Moby issue tracker);
> 
> * [RFE: Reworking Seccomp Profiles to ENOSYS Default moby#42871](https://github.com/moby/moby/issues/42871)
> * possibly we could use the work that was done in `containers/common` (most notably [seccomp: switch default to ENOSYS containers/common#573](https://github.com/containers/common/pull/573))
> 
> ☝️ whatever we do on that matter, we should try to preserve git history for all of those changes, as there's various cases where history contains very relevant information (why was a `syscall` added / removed? what was the motivation and decision around it?)

Hmm, isn't it solved in runc (https://github.com/opencontainers/runc/pull/2750, runc-1.0.0-rc93) already? I _think_ this is also fixed in crun but can't find a PR at the moment.

More to say, there's also an on-going work to implement it in libseccomp: https://github.com/seccomp/libseccomp/pull/457

PR #190, opened 2025-06-23
reexec: only consider basename, not path

thaJeztah:
The intent of reexec is to provide functionality similar to multi-call binaries (like "busybox"). However, currently, re-exec also includes path information when matching. This does not match behavior like busybox;

    date --help 2>&1 | grep Usage
    Usage: date [OPTIONS] [+FMT] [[-s] TIME]

    mkdir foo
    ln -s /bin/busybox ./foo/date
    ./foo/date --help 2>&1 | grep Usage

This patch changes the behavior to only consider the basename, and disregard path.

With this patch applied:

Basic example:

```go
package main

import (
	"fmt"
	"os"

	"github.com/moby/sys/reexec"
)

func init() {
	reexec.Register("hello-foo", func() {
		fmt.Println("Hello Foo")
		os.Exit(0)
	})
	reexec.Register("hello-bar", func() {
		fmt.Println("Hello Bar")
		os.Exit(0)
	})

	reexec.Init()
}

func main() {
	fmt.Println("Hello Main")
}
```

executing the binary, including path (`./`) now selects the correct
function to execute.

```bash
go build -o ./example
ln -s ./example hello-foo
ln -s ./example hello-bar
ln -s ./example hello-baz

./example
# Hello Main

./hello-foo
# Hello Foo

./hello-bar
# Hello Bar

./hello-baz
# Hello Main
```

Before this patch, all of the above would fall through to use `main` instead;

```bash
./example
# Hello Main

./hello-foo
# Hello Main

./hello-bar
# Hello Main

./hello-baz
# Hello Main
```

### DIFF
--- a/reexec/reexec.go
+++ b/reexec/reexec.go
@@ -19,6 +19,9 @@ var registeredInitializers = make(map[string]func())
 // Register adds an initialization func under the specified name. It panics
 // if the given name is already registered.
 func Register(name string, initializer func()) {
+	if filepath.Base(name) != name {
+		panic(fmt.Sprintf("reexec func does not expect a path component: %q", name))
+	}
 	if _, exists := registeredInitializers[name]; exists {
 		panic(fmt.Sprintf("reexec func already registered under name %q", name))
 	}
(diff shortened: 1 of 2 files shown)


thaJeztah:
I'll bring this one in; did a quick check, and I didn't find any place where we call `Register` with a command that includes a path (but will open draft PR's before doing a new release - some other changes still coming)

PR #193, opened 2025-06-24
reexec: add "CommandContext"

thaJeztah:
- follow-up to / stacked on https://github.com/moby/sys/pull/191

Adds a CommandContext command, to provide the equivalent of exec.CommandContext.

PR #194, opened 2025-06-24
reexec: test cmd.Cancel, bump minimum go version to go1.20

thaJeztah:
- stacked on https://github.com/moby/sys/pull/193

---

### reexec: test cmd.Cancel, bump minimum go version to go1.20

This requires go1.20 or up; https://go-review.googlesource.com/23300

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,8 +25,9 @@ jobs:
       if: ${{ matrix.go-version == '1.18.x' }}
       run: |
         # This corresponds with the list in Makefile:1, but omits the "userns"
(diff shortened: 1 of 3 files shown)


kolyshkin (approved):
LGTM

(I dream about the day we can drop Go 1.18 entirely, and just use supported versions)

thaJeztah:
Yeah, it's a bit of extra hassle, then again for modules that don't need it, it reduces having to update those. Perhaps we should check the `go.mod` for version, and based on that skip the module in the matrix (so that we don't have to maintain a list).

PR #206, opened 2026-03-01
reexec: touch-up docs and add some example uses

thaJeztah:


reexec/reexec.go
```diff
@@ -1,9 +1,72 @@
...
+//     and argv[0] (see [execve(2)]) selects the entrypoint.
+//   - For programmatic re-exec: a parent launches the current binary (via
+//     [Command] or [CommandContext]) with argv[0] set to a registered entrypoint
+//     name to run a specific mode.
+//
+// The programmatic re-exec pattern is commonly used as a safe alternative to
+// fork-without-exec in Go, since the runtime is not fork-safe once multiple
+// threads exist (see [os.StartProcess] and https://go.dev/issue/27505).
+//
+// Example (multi-call binary):
```
kolyshkin, line 25: nit: I would move example to a separate file, example_test.go or so. Having example in comments means that it is never even compiled, which will make it outdated.

Go examples are at least compiled.

thaJeztah: Yeah; I did that for the other one; but was looking for a "here's how it would be used for multi-call binary", and of course the symlinks aren't really something to do in an example test.

kolyshkin: > and of course the symlinks aren't really something to do in an example test.

I think it can still be done, you just create the binary first using `go test -c` and then go create symlinks. I haven't tested this, but in case your example has `// Output: ` line at the end of the function, it will actually be run during tests (to check the output matches).

I agree it can be cumbersome though, so if you want it to leave as it, I'm fine with it.

thaJeztah: Yeah, it's possible to test probably, but I don't think that works for `pkg.go.dev` to do something similar to that 🤔 

I think this one is fine to keep; I'm adding a second "mocked" example for this scenario in https://github.com/moby/sys/pull/208

vvoland (approved):

PR #207, opened 2026-03-01
reexec: improve test-coverage, and use blackbox testing

thaJeztah:


reexec/internal/reexecoverride/reexecoverride.go
```diff
@@ -0,0 +1,47 @@
...
+		return "", false
+	}
+	return *p, true
+}
+
+// TestingTB is the minimal subset of [testing.TB] used by this package.
+type TestingTB interface {
+	Helper()
+	Cleanup(func())
+}
```
thaJeztah, line 24: I went for a local interface after all; https://github.com/moby/sys/pull/208#discussion_r2874283089 because this package is imported in `reexec.go` ("production" code)

reexec/internal/reexecoverride/reexecoverride.go
```diff
@@ -0,0 +1,47 @@
...
+	if !argv0Override.CompareAndSwap(nil, &s) {
+		panic("testing: test using reexecoverride.OverrideArgv0 cannot use t.Parallel")
+	}
+
+	t.Cleanup(func() {
+		if !argv0Override.CompareAndSwap(&s, nil) {
+			panic("testing: cleanup for reexecoverride.OverrideArgv0 detected parallel use of reexec.Self")
+		}
+	})
+}
```
thaJeztah, line 45: Implemented your suggestion here, @kolyshkin 👍

reexec/reexec.go
```diff
@@ -78,14 +80,18 @@ func CommandContext(ctx context.Context, args ...string) *exec.Cmd {
 // "my-binary" at "/usr/bin/" (or "my-binary.exe" at "C:\" on Windows),
 // then it returns "/usr/bin/my-binary" and "C:\my-binary.exe" respectively.
 func Self() string {
+	if argv0, ok := reexecoverride.Argv0(); ok {
+		return naiveSelf(argv0)
+	}
 	if runtime.GOOS == "linux" {
 		return "/proc/self/exe"
 	}
```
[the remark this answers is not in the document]
thaJeztah: Nah, not worth documenting publicly I think

reexec/reexec_test.go
```diff
@@ -194,30 +200,88 @@ func TestCommandContext(t *testing.T) {
...
+			executable = "cmd"
+		}
+		reexecoverride.OverrideArgv0(t, executable)
+		resolved := reexec.Self()
+		if resolved == executable {
+			t.Errorf("did not resolve via PATH; got %q", resolved)
+		}
+		if !filepath.IsAbs(resolved) {
+			t.Errorf("expected absolute path; got %q", resolved)
+		}
```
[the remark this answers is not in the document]
thaJeztah: `exec.LookPath` should return an absolute path, no? @copilot code review[agent] 

```go
// LookPath searches for an executable named file in the current path,
// following the conventions of the host operating system.
// If file contains a slash, it is tried directly and the default path is not consulted.
// Otherwise, on success the result is an absolute path.
//
// LookPath returns an error satisfying [errors.Is](err, [ErrDot])
// if the resolved path is relative to the current directory.
// See the package documentation for more details.
```

reexec/internal/reexecoverride/reexecoverride.go
```diff
@@ -0,0 +1,45 @@
+// Package reexecoverride provides test utilities for overriding argv0 as
+// observed by reexec.Self within the current process.
+
```
vvoland, line 3: Shouldn't this be in a `_test.go` file?

thaJeztah: No, because you can't import `_test.go` files from another package; in this case we need to inject a test-override in a non-test file, and from outside the package, so tried to get the best option to;

- allow the module to import it
- prevent it being imported by anything outside of the module (so `internal/`)

vvoland: Hmmm right

vvoland (approved):

PR #208, opened 2026-03-01
reexec: add runnable examples

thaJeztah:
- [x] depends on / stacked on https://github.com/moby/sys/pull/207

### DIFF
--- a/reexec/example_multicall_test.go
+++ b/reexec/example_multicall_test.go
@@ -1,0 +1,59 @@
+package reexec_test
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/moby/sys/reexec"
+)
+
+func init() {
+	reexec.Register("example-foo", func() {
+		fmt.Println("Hello from entrypoint example-foo")
+	})
+	reexec.Register("example-bar", func() {
+		fmt.Println("Hello from entrypoint example-bar")
+	})
+}
+
+// Example_multicall demonstrates a BusyBox-style multi-call binary.
+//
+// In a real multi-call binary:
+//
+//	go build -o example .
+//	ln -s example example-foo
+//	ln -s example example-bar
+//
+//	./example-foo  # runs entrypoint "example-foo"
+//	./example-bar  # runs entrypoint "example-bar"
+//
+// At process startup, main would call [reexec.Init] and return if it
+// matches an entrypoint. This example first shows that call, then emulates
+// different invocation names by modifying os.Args[0].
+func Example_multicall() {
+	// What main would normally do:
+	if reexec.Init() {
+		// Matched a reexec entrypoint; stop normal main execution.
+		return
+	}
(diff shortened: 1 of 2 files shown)


reexec/internal/reexectest/reexectest.go
```diff
@@ -0,0 +1,42 @@
...
+	return *p, true
+}
+
+// OverrideArgv0 overrides os.Arg[0] with argv0 and returns a restore function.
+// Call the restore function (typically via t.Cleanup) to clear the override.
+//
+// While the override is active, any concurrent call to reexec.Self() in the
+// same process will observe the override. So tests that use this should not
+// run in parallel with other tests that call reexec.Self().
+func OverrideArgv0(argv0 string) (restore func()) {
```
kolyshkin: Since this is for testing, I'd modify it to accept testing.TB and using tb.Cleanup, rather than returning a cleanup function.

reexec/internal/reexectest/reexectest.go
```diff
@@ -0,0 +1,42 @@
...
+	overrideMu.Lock()
+	s := argv0
+	argv0Override.Store(&s)
```
kolyshkin: Using mutex to avoid tests being executed in parallel seems problematic here (tests might timeout or exhibit weird behavior).

Perhaps it's better to do something like this instead (not tested):

```go
if argv0Override.CompareAndSwap(&s, nil) {
  panic("tests using reexec should not be run in parallel")
}
t.Cleanup(func() {
  if argv0Override.CompareAndSwap(nil, &s) {
    panic("cleanup: tests using reexec should not be run in parallel")
  }
})
```

This way, as long as tests are executed sequentially, everything is fine. If tests are executed in parallel, they will panic.

reexec/internal/reexectest/reexectest.go
thaJeztah: oh! maybe that works yes; I was trying to find an elegant solution to do black box testing 😅

thaJeztah: Not sure if there's downsides to importing "testing" in production code? Probably should be fine?

thaJeztah: Guess I could still define an a local interface here for it 🤔

kolyshkin: > Not sure if there's downsides to importing "testing" in production code? Probably should be fine?

It's an internal package only used by tests, so why not

reexec/internal/reexectest/reexectest.go
thaJeztah: Ah, yes, I like that; I was originally contemplating to abuse `t.SetEnv` (or similar) for it, because it has detection for parallel 😂 (I wish that was exposed by `testing`, e.g. a `testing.IsParallel()`)

vvoland (approved):

PR #212, opened 2026-03-04
devices: move libcontainer/devices to moby/sys

### LINKED ISSUE #181: proposal: add
There is a set of functions (`DeviceFromPath`, `HostDevices`, `GetDevices`) which are copied across a few repos, in some cases with minor changes:
 
* https://github.com/opencontainers/runc/blob/main/libcontainer/devices/device_unix.go (not used by runc itself, used by containers/common, containers/buildah, cri-o, microsoft/hcsshim);
* https://github.com/containerd/containerd/blob/main/pkg/oci/utils_unix.go (used mostly from containerd and moby)
* https://github.com/containers/podman/blob/main/pkg/util/utils_linux.go (used from containers/*)

So, main users are moby/moby, cri-o/cri-o, containerd, faasd, microsoft/hcsshim, containers/*, etc.

I propose to create `moby/sys/devices`, figure out the implementation which will suit all the users, and add it there.

cyphar:
This is migrated from runc's libcontainer/devices, as this package has
only ever really been used outside of runc to generate device
configurations, so it makes more sense for it to live in
github.com/moby/sys.

With regards to the go.mod used, github.com/opencontainers/cgroups
requires Go 1.24, but aside from that all of these requirements are as
minimal as possible.

This merge was generated with `git filter-repo --path-glob="libcontainer/devices/*" --path-rename=libcontainer/devices:devices`. This does lose some of the pre-2015 history (opencontainers/runc@8f97d39dd250418fdbc9bc266bfdcbf9fd5a420b moved `libcontainer` into a subdirectory which means that the `libcontainer/devices` glob is missing commits before then) but this should be sufficient...

Fixes #181


devices/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/moby/sys/devices
+
+go 1.24
+
+require (
+	github.com/opencontainers/cgroups v0.0.6
+	golang.org/x/sys v0.41.0
```
kolyshkin: This is probably unnecessary, and imposes an unnecessary requirement for users of this package (to have a very new x/sys).

All other packages in this repo have golang.org/x/sys v0.1.0 in their go.mod files.

devices/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/moby/sys/devices
+
+go 1.24
```
kolyshkin, line 3: Also, here we'd like to be more conservative. If possible, use `go 1.18`. If not, the earliest version that works (no need to "downgrade" the code I guess).

kolyshkin: Ah! It might be that oc/cgroups v0.0.6 require go 1.24 already. If that's the case, so be it.

cyphar: Yeah that was the reason.

kolyshkin (approved):
LGTM

@thaJeztah PTAL

kolyshkin:
@thaJeztah PTAL 🙏🏻

kolyshkin:
@thaJeztah PTAL 🙏🏻

thaJeztah:
WELP #toomanynotifications sorry! Let me have a peek 🤗

thaJeztah (approved):
looks like a clean move


LGTM, thanks!

thaJeztah:
@kolyshkin want to do the honours of tagging ? Not sure what version to start with but v0.1.0 probably would do the trick?

kolyshkin:
> @kolyshkin want to do the honours of tagging ? Not sure what version to start with but v0.1.0 probably would do the trick?

Done: 
https://github.com/moby/sys/releases/tag/devices%2Fv0.1.0

PR #213, opened 2026-03-25
ci: pin actions by sha

thaJeztah:


kolyshkin:
To me this (using sha or defining a complete specific version) feels like a totally unnecessary complication which will create extra noise (more frequent updates, etc.).

Unless, of course,
  - a specific action used tends to break often;
  - or GHA is used for something very important and security-sensitive, like building an official release binary.
  
 Neither seems to be the case here.

kolyshkin:
OTOH looks like everyone else is doing it these days, so why shouldn't we.

kolyshkin (approved):

thaJeztah:
Yeah, I hate it; I think this should've been something that would be solved through something similar to Go's checksumdb and/or an option to point to a custom DB (with an allow/deny-list of versions / checksums).


I'll bring this one in.

PR #214, opened 2026-03-26
user: add filesystem-aware mkdir/chown helpers

tonistiigi:
Add FS-backed MkdirAllAndChownFS and MkdirAndChownFS variants for better security via os.Root implementation.

### DIFF
--- a/user/idtools.go
+++ b/user/idtools.go
@@ -3,6 +3,21 @@ package user
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+)
+
+// FS is the filesystem contract used by the Mkdir*AndChownFS helpers.
+type FS interface {
+	Stat(name string) (os.FileInfo, error)
+	Mkdir(name string, perm os.FileMode) error
+	MkdirAll(name string, perm os.FileMode) error
+	Chmod(name string, mode os.FileMode) error
+	Chown(name string, uid, gid int) error
+}
+
+var (
+	_ FS = &os.Root{}
+	_ FS = &hostFS{}
 )
 
 // MkdirOpt is a type for options to pass to Mkdir calls
@@ -23,12 +38,24 @@ func WithOnlyNew(o *mkdirOptions) {
 // function will still change ownership and permissions. If WithOnlyNew is passed as an
 // option, then only the newly created directories will have ownership and permissions changed.
 func MkdirAllAndChown(path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
-	var options mkdirOptions
-	for _, opt := range opts {
-		opt(&options)
+	return MkdirAllAndChownFS(nil, path, mode, uid, gid, opts...)
+}
+
+// MkdirAllAndChownFS creates a directory (including any along the path) on the
+// provided filesystem and then modifies ownership to the requested uid/gid. If
+// fsys is nil, the host filesystem is used.
+func MkdirAllAndChownFS(fsys FS, path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
+	if fsys == nil {
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return err
+		}
+		fsys = &hostFS{}
+		path = absPath
 	}
 
-	return mkdirAs(path, mode, uid, gid, true, options.onlyNew)
+	options := mkdirOpts(opts)
+	return mkdirAs(fsys, path, mode, uid, gid, true, options.onlyNew)
 }
 
 // MkdirAndChown creates a directory and then modifies ownership to the requested uid/gid.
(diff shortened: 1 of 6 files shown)


tonistiigi:
Technically `os.Root` dependency is not needed for other that tests, but no need to spend time on maintaining 1.18 support.

kolyshkin:
> no need to spend time on maintaining 1.18 support.

If we are to drop go 1.18 support (btw I'm all for it!), we need to
 - do it in a separate PR;
 - bump `go` version in `go.mod` for all modules;
 - apply modernize and clean old stuff for all modules;
 - remove the CI kludges for older Go versions (this is what the second commit here currently does).
 
 @thaJeztah WDYT?
 
Or, we can bump `go` version in `user/go.mod` only and everything should work.

tonistiigi:
> Or, we can bump go version in user/go.mod only and everything should work.

That's what the PR is doing atm.

kolyshkin:
> > Or, we can bump go version in user/go.mod only and everything should work.
> 
> That's what the PR is doing atm.

it looks like you're also dropping go 1.18 from CI (or is this a leftover?)

thaJeztah:
Yeah, no problem dropping go1.18 for this module; we kept the module versions at the minimum required if there was no reason we needed a newer version.

But you can leave the go1.18 version in CI; Ci will skip it if a module doesn't support it (but run it for the other modules that do).

user/idtools.go
```diff
@@ -23,12 +38,24 @@ func WithOnlyNew(o *mkdirOptions) {
...
-	var options mkdirOptions
-	for _, opt := range opts {
-		opt(&options)
+	return MkdirAllAndChownFS(nil, path, mode, uid, gid, opts...)
+}
+
+// MkdirAllAndChownFS creates a directory (including any along the path) on the
+// provided filesystem and then modifies ownership to the requested uid/gid. If
+// fsys is nil, the host filesystem is used.
+func MkdirAllAndChownFS(fsys FS, path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
```
thaJeztah, line 47: Looking at the signature; I see the variadic options at the end; did we add those later (but kept the positional options because they already were there?)

If that's the case, and we want to make more of these "optional", then now is the opportunity to change the signature for the new functions we added.

(also looking at windows <--> linux w.r.t. "uid", "gid" etc)

user/idtools.go
```diff
@@ -23,12 +38,24 @@ func WithOnlyNew(o *mkdirOptions) {
 // function will still change ownership and permissions. If WithOnlyNew is passed as an
 // option, then only the newly created directories will have ownership and permissions changed.
 func MkdirAllAndChown(path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
```
thaJeztah, line 40: if we want users of this module to migrate to the new functions, we can add a `//go:fix inline`, then `go fix` will automatically migrate them 🎉 


```suggestion
// option, then only the newly created directories will have ownership and permissions changed.
//
//go:fix inline
func MkdirAllAndChown(path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
```

user/idtools.go
```diff
@@ -38,11 +65,32 @@ func MkdirAllAndChown(path string, mode os.FileMode, uid, gid int, opts ...Mkdir
 // Note that unlike os.Mkdir(), this function does not return IsExist error
 // in case path already exists.
 func MkdirAndChown(path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
```
thaJeztah, line 67: Same for this one;

```suggestion
// in case path already exists.
//
//go:fix inline
func MkdirAndChown(path string, mode os.FileMode, uid, gid int, opts ...MkdirOpt) error {
```

user/idtools.go
kolyshkin: In this case (if you want everyone to migrate) it also makes sense to mark old ones as deprecated.

OTOH if you want the new functions signatures to be simpler as you're suggesting above, the go:fix won't be possible (or easy).

thaJeztah: > it also makes sense to mark old ones as deprecated.

Yup, deprecating is an option for sure; didn't want to make that a strict requiremnt for this PR (good as a follow-up, also for visibility).

> OTOH if you want the new functions signatures to be simpler as you're suggesting above,

I'd be fine with keeping the signature the same, but at least wanted to mention the option, because now we have the chance to do so.

Specifically in these cases where Golang tries to abstract away platform differences, and on Windows making (e.g.) `Chown` and `Chmod` (for permissions) a no-op, silently discarding options that the user may have expected to secure things.

Arguments that are required should probably still be positional though (which is the fun thing here if those are required on Linux, but won't take effect on Windows I guess); happy for input on that though!

> the go:fix won't be possible (or easy).

Yeah, `//go:fix` is not a MUST, but can help. It can do a pretty decent job; erm, but of course when I was about to write an example, I found yet another corner-case :joy: - adding it to my list of reports; https://github.com/golang/go/issues/78432

tonistiigi:
I only need a way to use these helpers without worrying about cross-system path access security. I have no opinion on deprecation of the old API or variadic options.

user/idtools.go
tonistiigi: Given that `Chown` is in the function name, I don't think `uid/gid` should be optional.

thompson-shaun:
@thaJeztah are we OK with the PR as it sits? looks like the feedback was largely optional?

thaJeztah:
It's "optional" as in; we won't be able to change after this, unless we rename the module, do a major release, or do trickery like `/v2` packages within a `v0` module.

So someone should do a quick check to see what arguments are absolutely required (cross-platform), and which arguments optional (or cannot be described cross-platform).

My comment on Windows was to avoid the same situation as go stdlib ended up with, which is that they silently shove security issues under the rug (just doesn't work on Windows), or that we end up with weird calls like;


Calling this is a complete no-op on Windows;

```go
MkdirAllAndChown("/must-be-secured", 0o700, 123, 456)
``` 

So having to do something like this;

```go
MkdirAllAndChown("/must-be-secured", 0o700, 123, 456, WithOwner("windows owner"), WithPermissions("windows permissions")) 
```

Because not setting those options would mean that permissions on Windows could potentially be wide-open.

PR #215, opened 2026-03-26
ci: test against per-module minimum go version

thaJeztah:
Go-setup action allow specifying both a "go-version" and "go-version-file", in which case it falls back to the version-file if "go-version" is empty.

This allows us to iterate over all modules, and run CI on the oldest supported Go version for each (in addition to "oldstable" and "stable").

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,12 +17,33 @@ on: [push, pull_request]
 
 jobs:
   test:
+    name: ${{ matrix.module }} / ${{ matrix.platform }} / ${{ matrix.go-version || 'minimum' }}
     strategy:
+      fail-fast: false
       matrix:
-        go-version: [1.18.x, oldstable, stable]
+        module:
+          - atomicwriter
+          - capability
+          - mount
+          - mountinfo
+          - reexec
+          - sequential
+          - signal
+          - symlink
+          - user
+          - userns
+        go-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -17,12 +17,33 @@ on: [push, pull_request]
...
+        go-version:
+          - "" # do not specify go-version to fall back on go-version-file (use go.mod); see https://github.com/actions/setup-go/issues/450#issuecomment-3620402646
+          - oldstable
+          - stable
         platform: [ubuntu-22.04, ubuntu-24.04, windows-2022, windows-2025, macos-15, macos-26]
     runs-on: ${{ matrix.platform }}
     timeout-minutes: 10 # guardrails timeout for the whole job
+
+    env:
+      GOTOOLCHAIN: local
```
kolyshkin, line 44: Is this a request for go to NOT download another go version? If yes, why do we need it here? I mean, it may only happen if we have go.mod with `go` version less than specified here, which should never be the case.

Or is this something more generic, like "don't download anything" and omitting it may mask some issue?

.github/workflows/test.yml
```diff
@@ -17,12 +17,33 @@ on: [push, pull_request]
...
+          - atomicwriter
+          - capability
+          - mount
+          - mountinfo
+          - reexec
+          - sequential
+          - signal
+          - symlink
+          - user
+          - userns
```
kolyshkin, line 34: I like the way it is used here. I don't like that we have to also list all the modules in the top-level Makefile.

PR #216, opened 2026-03-31
sequential: use os.OpenFile with O_FILE_FLAG_SEQUENTIAL_SCAN on Go 1.26+

tonistiigi:
Go 1.26 adds support for passing Windows file flags via os.OpenFile, eliminating the need to call windows.CreateFile directly. Split the implementation into build-tagged files: go1.26+ uses os.OpenFile with the new O_FILE_FLAG_SEQUENTIAL_SCAN open flag, while older versions retain the manual CreateFile approach.

Also bump golang.org/x/sys to v0.37.0 (provides the new constant) and update the minimum Go version to 1.24.

https://go-review.googlesource.com/c/go/+/699415
https://go-review.googlesource.com/c/go/+/724621

### DIFF
--- a/sequential/sequential_windows_go126.go
+++ b/sequential/sequential_windows_go126.go
@@ -1,0 +1,13 @@
+//go:build windows && go1.26
+
+package sequential
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+func openFileSequential(name string, flag int, perm os.FileMode) (*os.File, error) {
+	return os.OpenFile(name, flag|windows.O_FILE_FLAG_SEQUENTIAL_SCAN, perm)
+}
--- a/sequential/go.mod
+++ b/sequential/go.mod
@@ -1,5 +1,5 @@
 module github.com/moby/sys/sequential
 
-go 1.18
+go 1.24.0
 
-require golang.org/x/sys v0.1.0
+require golang.org/x/sys v0.37.0
--- a/sequential/go.sum
+++ b/sequential/go.sum
@@ -1,2 +1,2 @@
-golang.org/x/sys v0.1.0 h1:kunALQeHf1/185U1i0GOB/fy1IPRDDpuoOOqRReG57U=
-golang.org/x/sys v0.1.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.37.0 h1:fdNQudmxPjkdUTPnLn5mdQv7Zwvbvpaxqs831goi9kQ=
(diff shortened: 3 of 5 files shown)


thaJeztah:
```console
# golang.org/x/sys/windows
Error: C:\Users\runneradmin\go\pkg\mod\golang.org\x\sys@v0.37.0\windows\security_windows.go:1318:39: undefined: unsafe.SliceData
Error: C:\Users\runneradmin\go\pkg\mod\golang.org\x\sys@v0.37.0\windows\security_windows.go:1321:39: undefined: unsafe.SliceData
Error: C:\Users\runneradmin\go\pkg\mod\golang.org\x\sys@v0.37.0\windows\security_windows.go:1324:40: undefined: unsafe.SliceData
Error: C:\Users\runneradmin\go\pkg\mod\golang.org\x\sys@v0.37.0\windows\security_windows.go:1327:40: undefined: unsafe.SliceData
```

Looks like we need to bump the module to go1.20 as minimum;

Screenshot 2026-04-01 at 23 38 23

thaJeztah:
Oh! Probably the other module that is (`moby/sys/mount`), which is tested against both the version specified in `go.mod` and against latest in the branch; 

```
MINGW64_NT-10.0-20348 runnervm6wf0k 3.6.6-1cdd4371.x86_64 2026-01-15 22:20 UTC x86_64 Msys
/usr/bin/sh: line 1: sudo: command not found
/usr/bin/sh: line 1: sudo: command not found
echo 'replace github.com/moby/sys/mountinfo => ../mountinfo' | cat mount/go.mod - > mount/go-local.mod
```

tonistiigi:
> Looks like we need to bump the module to go1.20 as minimum;

I had already bumped to 1.24 (via x/sys update that had this as minimum). This is the 1.18 hacks.

I removed my previous attempt to patch the 1.18 workarounds and removed it from CI like I initially did in #214 . I've already spent way too much time on these 1.18 exceptions that nobody needs.

crazy-max:
image

We should remove the required checks in branch protection. Makes little sense to keep them anyway.

thaJeztah (approved):
LGTM

sequential/sequential_windows_go126.go
```diff
@@ -0,0 +1,13 @@
...
+package sequential
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+func openFileSequential(name string, flag int, perm os.FileMode) (*os.File, error) {
+	return os.OpenFile(name, flag|windows.O_FILE_FLAG_SEQUENTIAL_SCAN, perm)
```
thaJeztah, line 12: once we drop support for < go1.26, we should define this as a local const, so that we no longer have a dependency on "golang.org/x/sys/windows"

PR #218, opened 2026-04-17
reexec: add context-aware entrypoints and Dispatch

thaJeztah:
- follow-up to https://github.com/moby/sys/pull/193

Follow-up to 7aa8514cb7058430152f19bdb002b6f02e183e92; allow entrypoints to be invoked with a context and to return an error;

- Add RegisterContext to register context-aware entrypoints.
- Add Dispatch and DispatchContext, both returning (handled, error).
- Init is kept for backward compatibility or cases where context or errors don't have to be handled.

### DIFF
--- a/reexec/example_programmatic_test.go
+++ b/reexec/example_programmatic_test.go
@@ -10,20 +10,26 @@ import (
 )
 
 func init() {
-	reexec.Register("example-child", func() {
+	reexec.RegisterContext("example-child", func(context.Context) error {
 		fmt.Println("Hello from example-child entrypoint")
+		return nil
 	})
 }
 
 // Example_programmatic demonstrates using reexec to programmatically
 // re-execute the current binary.
 func Example_programmatic() {
-	if reexec.Init() {
-		// Matched a reexec entrypoint; stop normal main execution.
+	ctx := context.Background()
+	if ok, err := reexec.Dispatch(ctx); ok {
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "entrypoint failed:", err)
+			os.Exit(1)
+		}
+		// Matched a reexec entrypoint; stop normal execution.
 		return
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
 	cmd := reexec.CommandContext(ctx, "example-child")
--- a/reexec/reexec.go
+++ b/reexec/reexec.go
@@ -80,15 +80,32 @@ import (
 	"github.com/moby/sys/reexec/internal/reexecoverride"
 )
 
-var entrypoints = make(map[string]func())
+var entrypoints = make(map[string]func(context.Context) error)
 
 // Register associates name with an entrypoint function to be executed when
 // the current binary is invoked with argv[0] equal to name.
 //
 // Register is not safe for concurrent use; entrypoints must be registered
-// during program initialization, before calling [Init].
+// during program initialization, before calling [Dispatch] or [Init].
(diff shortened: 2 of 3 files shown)


reexec/reexec_test.go
```diff
@@ -22,35 +23,66 @@ const (
...
 	})
 	reexec.Register(testReExec3, func() {
 		fmt.Println("Hello " + testReExec3)
 		time.Sleep(1 * time.Second)
 		os.Exit(0)
 	})
-	if reexec.Init() {
-		// Make sure we exit in case re-exec didn't os.Exit on its own.
+
+	if ok, err := reexec.DispatchContext(context.Background()); ok {
```
thaJeztah, line 46: oh! this one was outdated; let me update to use t.Context

thaJeztah: oh, nevermind; this is not in a test

reexec/reexec.go
```diff
@@ -99,18 +116,45 @@ func Register(name string, entrypoint func()) {
...
+// match with ok. If ok is false, err is always nil. If ok is true, err is the
+// entrypoint's return value.
+//
+// In the ok=true case, the caller should stop normal main execution (typically
+// by returning from main or calling os.Exit).
+func Dispatch() (bool, error) {
+	return dispatch(context.Background())
+}
+
+// DispatchContext is like [Dispatch] but includes a context.
```
kolyshkin: Since it's a new function, maybe
 - remove context-less version entirely;
 - rename the context-ful version to `Dispatch`.

reexec/reexec_test.go
```diff
@@ -22,35 +23,66 @@ const (
...
+		if err != nil {
+			_, _ = fmt.Fprintln(os.Stderr, "entrypoint failed:", err)
```
kolyshkin, line 48: nit: is `_, _ = ` really needed here?

thaJeztah: Mostly trying to keep my IDE happy; I think linters are more relaxed, but not all of them may be, so I generally prefer the code itself to be explicit over adding `//nolint` (or similar) comments.

reexec/reexec.go
thaJeztah: Yes, I guess that makes sense; we'd still have `Init` as alternative, so yeah.

thaJeztah:
GitHub having a bad day? Looks like possibly networking issues

thaJeztah:
All green now; updated; PTAL 🤗

vvoland (approved):

reexec/example_programmatic_test.go
```diff
@@ -10,15 +10,20 @@ import (
 )
 
 func init() {
-	reexec.Register("example-child", func() {
+	reexec.RegisterContext("example-child", func(ctx context.Context) error {
```
kolyshkin: very minor nit: omit `ctx` as you don't use this variable.

reexec/reexec_test.go
kolyshkin: Not very important, but I'd remove it for less noise and consistency with other few places (for example, the one you add to reexec/example_programmatic_test.go in the same commit).

kolyshkin (approved):
two nits but not very important so LGTM (feel free to fix or merge as is)

reexec/reexec_test.go
thaJeztah: > for example, the one you add to reexec/example_programmatic_test.go in the same commit

Yeah, for that one, I was trying to avoid noise as part of the example shown as it's mostly for documentation purposes.

reexec/example_programmatic_test.go
thaJeztah: removed

PR #220, opened 2026-04-30
use NtSetInformationFile if os.Rename fails

karman-docker:
This PR adds `atomicwriterRenameAt` which uses `NtSetInformationFile` to rename a file on Windows.

### DIFF
--- a/atomicwriter/atomicwriter_unix.go
+++ b/atomicwriter/atomicwriter_unix.go
@@ -1,0 +1,9 @@
+//go:build !windows
+
+package atomicwriter
+
+import "os"
+
+func atomicwriterRename(oldpath, newpath string) error {
+	return os.Rename(oldpath, newpath)
+}
--- a/atomicwriter/atomicwriter_unix.go
+++ b/atomicwriter/atomicwriter_unix.go
@@ -4,6 +4,6 @@ package atomicwriter
 
 import "os"
 
-func atomicwriterRenameAt(oldpath, newpath string) error {
+func atomicwriterRename(oldpath, newpath string) error {
 	return os.Rename(oldpath, newpath)
 }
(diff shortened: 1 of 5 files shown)


atomicwriter/atomicwriter_unix.go
```diff
@@ -0,0 +1,9 @@
+//go:build !windows
+
+package atomicwriter
+
+import "os"
+
+func atomicwriterRenameAt(oldpath, newpath string) error {
```
kolyshkin: nit: stuttering

Also, why is it called rename **At**? Usually 'At' implies there's an dirfd provided, which is not the case.

kolyshkin:
This obviously needs some windows-specific tests.

kolyshkin (commented):
nit: you've introduced `atomicwriterRenameAt` in the first commit, and then rename it to `atomicwriterRename` in the second commit. Why not use the right name from the start?

nit: stuttering at the name (atomicwriter.atomicwriterRename).

karman-docker:
> nit: you've introduced `atomicwriterRenameAt` in the first commit, and then rename it to `atomicwriterRename` in the second commit. Why not use the right name from the start?
> 
> nit: stuttering at the name (atomicwriter.atomicwriterRename).

sorry I thought in your initial [review comment](https://github.com/moby/sys/pull/220#discussion_r3327047661), you suggested to remove 'At' ?

kolyshkin:
> > nit: you've introduced `atomicwriterRenameAt` in the first commit, and then rename it to `atomicwriterRename` in the second commit. Why not use the right name from the start?
> > nit: stuttering at the name (atomicwriter.atomicwriterRename).
> 
> sorry I thought in your initial [review comment](https://github.com/moby/sys/pull/220#discussion_r3327047661), you suggested to remove 'At' ?

Yes, but here I'm not saying `At` should be back, I am saying the right name should be introduced from the start, i.e. in the first commit.

PR #221, opened 2026-05-26
user: prevent falling back to looking up numeric usernames

thaJeztah:
- [x] depends on / stacked on https://github.com/moby/sys/pull/226
- relates to https://github.com/containerd/containerd/pull/13446
- relates to https://github.com/containerd/containerd/commit/de1341c201ffb0effebbf51d00376181968c8779

Addresses a similar issue as [CVE-2026-46680](https://github.com/containerd/containerd/security/advisories/GHSA-fqw6-gf59-qr4w) in containerd.

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -259,6 +259,38 @@ func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath 
 	return GetExecUser(userSpec, defaults, passwd, group)
 }
 
+// parseNumeric parses the given UID or GID value to an integer and within
+// the minID - maxID range.
+//
+// While the Linux kernel allows the max UID to be MaxUint32 - 2,
+// and the OCI Runtime Spec has no definition about the max UID, we require
+// the UID to be <= MaxInt32.
+//
+// See https://github.com/containerd/containerd/commit/de1341c201ffb0effebbf51d00376181968c8779
+func parseNumeric(val string) (int, bool, error) {
+	if val == "" {
+		return 0, false, nil
+	}
+	id, err := strconv.Atoi(val)
+	if err != nil {
+		if errors.Is(err, strconv.ErrSyntax) {
+			// Discard the error, because non-numeric values are expected
+			// when passing a username or group-name.
+			return 0, false, nil
+		}
+		if errors.Is(err, strconv.ErrRange) {
+			return 0, true, ErrRange
+		}
+		// Other errors ("invalid base", "invalid bit size"); should never
+		// happen with strconv.Atoi.
+		return 0, false, err
+	}
+	if id < minID || id > maxID {
+		return 0, true, ErrRange
+	}
+	return id, true, nil
+}
+
 // GetExecUser parses a user specification string (using the passwd and group
 // readers as sources for /etc/passwd and /etc/group data, respectively). In
 // the case of blank fields or missing data from the sources, the values in
@@ -304,8 +336,14 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 
 	// Convert userArg and groupArg to be numeric, so we don't have to execute
 	// Atoi *twice* for each iteration over lines.
-	uidArg, uidErr := strconv.Atoi(userArg)
-	gidArg, gidErr := strconv.Atoi(groupArg)
+	uidArg, isUID, err := parseNumeric(userArg)
+	if err != nil {
+		return nil, err
+	}
+	gidArg, isGID, err := parseNumeric(groupArg)
+	if err != nil {
+		return nil, err
+	}
 
 	// Find the matching user.
 	users, err := ParsePasswdFilter(passwd, func(u User) bool {
@@ -314,8 +352,8 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 			return u.Uid == user.Uid
 		}
 
-		if uidErr == nil {
-			// If the userArg is numeric, always treat it as a UID.
+		if isUID {
+			// If the userArg is a valid numeric value, always treat it as a UID.
 			return uidArg == u.Uid
 		}
 
@@ -341,18 +379,11 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 		// If we can't find a user with the given username, the only other valid
 		// option is if it's a numeric username with no associated entry in passwd.
 
-		if uidErr != nil {
+		if !isUID {
 			// Not numeric.
 			return nil, fmt.Errorf("unable to find user %s: %w", userArg, ErrNoPasswdEntries)
 		}
 		user.Uid = uidArg
-
-		// Must be inside valid uid range.
-		if user.Uid < minID || user.Uid > maxID {
-			return nil, ErrRange
-		}
-
-		// Okay, so it's numeric. We can just roll with this.
 	}
 
 	// On to the groups. If we matched a username, we need to do this because of
@@ -370,7 +401,7 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 				return false
 			}
 
-			if gidErr == nil {
+			if isGID {
 				// If the groupArg is numeric, always treat it as a GID.
 				return gidArg == g.Gid
 			}
@@ -390,18 +421,11 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 				// If we can't find a group with the given name, the only other valid
 				// option is if it's a numeric group name with no associated entry in group.
 
-				if gidErr != nil {
+				if !isGID {
 					// Not numeric.
 					return nil, fmt.Errorf("unable to find group %s: %w", groupArg, ErrNoGroupEntries)
 				}
 				user.Gid = gidArg
-
-				// Must be inside valid gid range.
-				if user.Gid < minID || user.Gid > maxID {
-					return nil, ErrRange
-				}
-
-				// Okay, so it's numeric. We can just roll with this.
 			}
 		} else if len(groups) > 0 {
 			// Supplementary group ids only make sense if in the implicit form.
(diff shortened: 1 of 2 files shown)


vvoland (commented):
This only tests the case where gid/uid still falls within `int` .

If it's larger, it will still resolve to root because it will be treated as an username and still resolved.

So:

```
9223372036854775808:x:0:0:root:/root:/bin/sh
```

+ `USER 9223372036854775808`

Means that the runtime will actually resolve it to `root`.

This is "correct" behavior, but its confusing to treat a numeric value as an username.

thaJeztah:
@vvoland - trying to see where that codepath would be; testing `GetExecUser` with `2147483648` (math.MaxInt32 + 1) produces an error; or is it a different codepath you're looking at?

vvoland:
On 64 bit `strconv.Atoi` will handle MaxInt32 + 1 without an error

vvoland:
Do we need the same for `GetAdditionalGroups`?

thaJeztah:
Yeah, makes sense; also updated


As a follow-up, I'm considering cleaning up this code; the logic seems to be much too complicated.

user/user.go
```diff
@@ -432,12 +461,34 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+
 	groups := []Group{}
 	if group != nil {
 		var err error
 		groups, err = ParseGroupFilter(group, func(g Group) bool {
-			for _, ag := range additionalGroups {
-				if g.Name == ag || strconv.Itoa(g.Gid) == ag {
+			for _, ag := range addtlGroups {
+				if ag.isNumeric {
+					return g.Gid == ag.gid
```
vvoland: Should this be

```suggestion
					if g.Gid == ag.gid {
						return true
					}
```

?

user/user.go
```diff
@@ -270,6 +270,43 @@ func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath
...
+		}
+		if r < '0' || r > '9' {
+			return 0, false, nil
+		}
+	}
+	if val == "+" || val == "-" {
+		return 0, false, nil
+	}
+
+	id, err := strconv.Atoi(val)
```
kolyshkin: I wonder why we don't use something like https://github.com/containerd/containerd/commit/85706b6d4416d93b47033ba345d7b885a75657b4 (i.e. use strconv.Atoi or strconv.ParseUint) and check its errors, as well as the range. Basically, here we parse the string twice, duplicating the strconv functionality.

user/user.go
```diff
@@ -270,6 +270,43 @@ func GetExecUserPath(userSpec string, defaults *ExecUser, passwdPath, groupPath
...
+// and the OCI Runtime Spec has no definition about the max UID, we require
+// the UID to be <= MaxInt32.
+//
+// See https://github.com/containerd/containerd/commit/de1341c201ffb0effebbf51d00376181968c8779
+func parseNumeric(val string) (int, bool, error) {
+	if val == "" {
+		return 0, false, nil
+	}
+	for i, r := range val {
+		if i == 0 && (r == '+' || r == '-') {
```
kolyshkin: I'm not sure that numeric UID/GID can legitimately start with + or -, but I have a strong feeling I am missing something here.

user/user.go
thaJeztah: No, don't think so, because then it would continue to check `if g.Name == ag.name {`, and we know it's a numeric value, so we can return the results (did it match, or not?)

user/user.go
thaJeztah: Ah, you're right; I was trying to detect early because the logic is slightly different here, but we can use the error returned here to detect whether it was a range error, or a syntax error (non-numeric).


Let me update

user/user.go
thaJeztah: Doh! You're right though; because we still want to continue the loop here 🙈

thaJeztah: Updated, and added a test-case;

```patch
diff --git a/user/user.go b/user/user.go
index 8012094..f47e250 100644
--- a/user/user.go
+++ b/user/user.go
@@ -480,8 +480,8 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
 		var err error
 		groups, err = ParseGroupFilter(group, func(g Group) bool {
 			for _, ag := range addtlGroups {
-				if ag.isNumeric {
-					return g.Gid == ag.gid
+				if ag.isNumeric && g.Gid == ag.gid {
+					return true
 				}
 				if g.Name == ag.name {
 					return true
diff --git a/user/user_test.go b/user/user_test.go
index 1e16861..6837e53 100644
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -477,6 +477,11 @@ this is just some garbage data
 			groups:   []string{"adm"},
 			expected: []int{43},
 		},
+		{
+			// numeric group miss must continue checking remaining groups
+			groups:   []string{"10001", "adm"},
+			expected: []int{43, 10001},
+		},
 		{
 			// multiple groups
 			groups:   []string{"adm", "grp"},
```

user/user.go
```diff
@@ -445,12 +445,35 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
 		var err error
 		groups, err = ParseGroupFilter(group, func(g Group) bool {
-			for _, ag := range additionalGroups {
-				if g.Name == ag || strconv.Itoa(g.Gid) == ag {
+			for _, ag := range addtlGroups {
+				if ag.isNumeric {
+					if g.Gid == ag.gid {
+						return true
+					}
+				} else if g.Name == ag.name {
```
kolyshkin: This logic repeats below, maybe it makes sense to isolate it for a slightly cleaner code?

```diff
diff --git a/user/user.go b/user/user.go
index 9142d21..fd6f6cb 100644
--- a/user/user.go
+++ b/user/user.go
@@ -464,16 +464,21 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
                }
        }
 
+       // matches reports whether group g satisfies the given argument. Numeric
+       // arguments are matched by GID only, others by name.
+       matches := func(g Group, ag groupArg) bool {
+               if ag.isNumeric {
+                       return g.Gid == ag.gid
+               }
+               return g.Name == ag.name
+       }
+
        groups := []Group{}
        if group != nil {
                var err error
                groups, err = ParseGroupFilter(group, func(g Group) bool {
                        for _, ag := range addtlGroups {
-                               if ag.isNumeric {
-                                       if g.Gid == ag.gid {
-                                               return true
-                                       }
-                               } else if g.Name == ag.name {
+                               if matches(g, ag) {
                                        return true
                                }
                        }
@@ -488,13 +493,8 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
        for _, ag := range addtlGroups {
                var found bool
                for _, g := range groups {
-                       // if we found a matched group either by name or gid, take the
-                       // first matched as correct
-                       if ag.isNumeric {
-                               if g.Gid != ag.gid {
-                                       continue
-                               }
-                       } else if g.Name != ag.name {
+                       // take the first matched group as correct
+                       if !matches(g, ag) {
                                continue
                        }
 
```

kolyshkin (approved):
LGTM overall; maybe just a slight improvement to the second commit?

user/user.go
thaJeztah: Ah, yes, that looks a bit cleaner.

(still an overall overhaul wouldn't hurt, but let's take it one step at a time)

kolyshkin: or we can have it as a method of groupArg:

```diff
diff --git a/user/user.go b/user/user.go
index 9142d21..a91a83f 100644
--- a/user/user.go
+++ b/user/user.go
@@ -439,18 +439,28 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
        return user, nil
 }
 
+// groupArg is a parsed group argument for [GetAdditionalGroups].
+type groupArg struct {
+       name      string
+       gid       int
+       isNumeric bool
+}
+
+// matches reports whether group g satisfies the argument. Numeric arguments
+// are matched by GID only, others by name.
+func (ag groupArg) matches(g Group) bool {
+       if ag.isNumeric {
+               return g.Gid == ag.gid
+       }
+       return g.Name == ag.name
+}
+
 // GetAdditionalGroups looks up a list of groups by name or group id
 // against the given /etc/group formatted data. If a group name cannot
 // be found, an error will be returned. If a group id cannot be found,
 // or the given group data is nil, the id will be returned as-is
 // provided it is in the legal range.
 func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, error) {
-       type groupArg struct {
-               name      string
-               gid       int
-               isNumeric bool
-       }
-
        addtlGroups := make([]groupArg, len(additionalGroups))
        for i, ag := range additionalGroups {
                gid, ok, err := parseNumeric(ag)
@@ -469,11 +479,7 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
                var err error
                groups, err = ParseGroupFilter(group, func(g Group) bool {
                        for _, ag := range addtlGroups {
-                               if ag.isNumeric {
-                                       if g.Gid == ag.gid {
-                                               return true
-                                       }
-                               } else if g.Name == ag.name {
+                               if ag.matches(g) {
                                        return true
                                }
                        }
@@ -488,13 +494,8 @@ func GetAdditionalGroups(additionalGroups []string, group io.Reader) ([]int, err
        for _, ag := range addtlGroups {
                var found bool
                for _, g := range groups {
-                       // if we found a matched group either by name or gid, take the
-                       // first matched as correct
-                       if ag.isNumeric {
-                               if g.Gid != ag.gid {
-                                       continue
-                               }
-                       } else if g.Name != ag.name {
+                       // take the first matched group as correct
+                       if !ag.matches(g) {
                                continue
                        }
 
```

thaJeztah: Hm, yes, I like that!

Considering for the `// take the first matched group as correct` block to flip the condition (I almost overlooked the `!`); now that it's much shorter, perhaps we should not use an early `continue`

kolyshkin (approved):
still lgtm

tonistiigi (approved):

PR #227, opened 2026-06-25
user: fix TestParseGroupFileCapsReads, skip TestParseGroupFilterDevZero on windows

thaJeztah:
This test broke after 2c56c3d5d089bfa66d01b6c0bb69428e9d5d18c5 added additional per-line constraints, which would be hit before the file- size limit was reached.

### DIFF
--- a/user/user_test.go
+++ b/user/user_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


vvoland (approved):

user/user_test.go
```diff
@@ -133,7 +133,13 @@ func TestParseGroupFileCapsReads(t *testing.T) {
 			tmpDir := t.TempDir()
 			fileName := filepath.Join(tmpDir, "etc-group")
 
-			data := append(bytes.Repeat([]byte{0}, tc.padBytes), beyond...)
+			pattern := []byte("# padding\n")
+			pad := bytes.Repeat(pattern, (tc.padBytes+len(pattern)-1)/len(pattern))[:tc.padBytes]
+			if len(pad) > 0 && pad[len(pad)-1] != '\n' {
```
thaJeztah, line 138: Maybe could still use closer to the original; let me try

```go
pad := bytes.Repeat([]byte{0}, tc.padBytes)
for i := 4095; i < len(pad); i += 4096 {
	pad[i] = '\n'
}

data := append(pad, beyond...)
```

thaJeztah: nevermind that won't work

PR #230, opened 2026-06-30
[WIP] user: skip malformed and empty lines

thaJeztah:


user/user.go
```diff
@@ -135,7 +154,11 @@ func ParsePasswdFilter(r io.Reader, filter func(User) bool) ([]User, error) {
 		//  root:x:0:0:root:/root:/bin/bash
 		//  adm:x:3:4:adm:/var/adm:/bin/false
 		p := User{}
-		parseLine(line, &p.Name, &p.Pass, &p.Uid, &p.Gid, &p.Gecos, &p.Home, &p.Shell)
+		ok, err := parseLine(line, &p.Name, &p.Pass, &p.Uid, &p.Gid, &p.Gecos, &p.Home, &p.Shell)
+		if err != nil || !ok {
+			// Skip malformed and empty lines, and don't consider them.
+			continue
+		}
```
thaJeztah, line 161: Possibly also skip zero-records;

```go
if p == (User{}) {
    continue
}
```

user/user.go
```diff
@@ -194,7 +217,11 @@ func ParseGroupFilter(r io.Reader, filter func(Group) bool) ([]Group, error) {
...
 		p := Group{}
-		parseLine(line, &p.Name, &p.Pass, &p.Gid, &p.List)
+		ok, err := parseLine(line, &p.Name, &p.Pass, &p.Gid, &p.List)
+		if err != nil || !ok {
+			// Skip malformed and empty lines, and don't consider them.
+			continue
+		}
 
 		if filter == nil || filter(p) {
 			out = append(out, p)
```
thaJeztah, line 227: Here as well, but `Group` is not comparable (due to the `List []string`)

user/user.go
```diff
@@ -135,7 +150,9 @@ func ParsePasswdFilter(r io.Reader, filter func(User) bool) ([]User, error) {
 		//  root:x:0:0:root:/root:/bin/bash
 		//  adm:x:3:4:adm:/var/adm:/bin/false
 		p := User{}
-		parseLine(line, &p.Name, &p.Pass, &p.Uid, &p.Gid, &p.Gecos, &p.Home, &p.Shell)
+		if err := parseLine(line, &p.Name, &p.Pass, &p.Uid, &p.Gid, &p.Gecos, &p.Home, &p.Shell); err != nil {
+			return nil, fmt.Errorf("invalid line (%q): %w", string(line), err)
```
kolyshkin: nit: you have both `()` and `""` around `line`, which seems excessive.
```suggestion
			return nil, fmt.Errorf("invalid line %q: %w", string(line), err)
```

user/user.go
```diff
@@ -194,7 +211,9 @@ func ParseGroupFilter(r io.Reader, filter func(Group) bool) ([]Group, error) {
 		//  root:x:0:root
 		//  adm:x:4:root,adm,daemon
 		p := Group{}
-		parseLine(line, &p.Name, &p.Pass, &p.Gid, &p.List)
+		if err := parseLine(line, &p.Name, &p.Pass, &p.Gid, &p.List); err != nil {
+			return nil, fmt.Errorf("invalid line (%q): %w", string(line), err)
```
kolyshkin: ditto

kolyshkin (changes requested):
The second patch seems problematic: while it says "skip malformed and empty lines", in fact it returns an error where previously it was ignored.

I think we should merely skip lines that do not conform to the specification, otherwise there will be regressions.

user/user.go
thaJeztah: Ah, yes, I think I was going back-and-forth on `%q` or `(%s)` - `%q` isn't always great due to escaping nested quotes if they can be expected.

thaJeztah:
> The second patch seems problematic: while it says "skip malformed and empty lines", in fact it returns an error where previously it was ignored.

Yes, I need to clean-up my patch set; temporarily pushed, because I think the second  (add error return) patch needed updates to the tests.


> I think we should merely skip lines that do not conform to the specification, otherwise there will be regressions.

This was the topic of discussion (so was reordering the patches to make the "skip" last);

Old code wasn't good, because it would silently discard invalid UID/GID (numeric) values; it would use `0` if parsing failed, and preserve malformed records. This would not always be problematic (i.e., if it's not the record we were looking for), but of course still collect records that are just baggage.

The trickier question is what to do if the file contains malformed line(s)?

Currently (see above) we "half" ignore them (bad). Which, even with this patch, may still be true;

We silently ignore truncated lines;

```go
// Ignore cases where we don't have enough fields to populate the arguments.
// Some configuration files like to misbehave.
if len(v) <= i {
    break
}
```

So a line like `adm:x:3` is still treated as an implicit `adm:x:3:0` (GID=0)

And we still consider empty fields;

```go
case *int64:
    if len(p) != 0 {
        n, err := strconv.ParseInt(string(p), 10, 64)
        if err != nil {
            return true, fmt.Errorf("parsing integer field %d: %w", i, err)
        }
        *e = n
    }
```

So `adm:x:::::::::::` Would be implicicly taken as `UID=0,GID=0` ?

The slightly less problematic case is ignoring completely malformed lines, and it's less problematic if such a record was not the one we were looking for (i.e., we wouldn't use it anyway), but of course, if it's malformed, it COULD have been the line we were looking for, and I think we pick "first match", so the malformed (or incoplete) line would shadow the valid one;

```
adm:x:3
adm:x:42:43:adm:/var/adm:/bin/false
```

If we skip such lines or error out, that would prevent that (but probably needs more than this patch).

thaJeztah:
Looks like getent skips over malformed lines;
https://sources.debian.org/src/glibc/2.42-13/nss/fgetpwent_r.c/
https://sources.debian.org/src/glibc/2.42-13/nss/nss_fgetent_r.c/#L46-L48

```c
  if (ret == EINVAL)
    /* Skip over malformed lines.  */
    continue;
```

Older implementations;
https://sources.debian.org/src/glibc/2.31-13%2Bdeb11u11/pwd/fgetpwent_r.c/#L100-L104


```c
    } while (*p == '\0' || *p == '#' /* Ignore empty and comment lines.  */
	     /* Parse the line.  If it is invalid, loop to
		get the next line of the file to parse.  */
	     || ! parse_line (p, resbuf, (void *) buffer, buflen, &errno));
```

But it looks like it does reject lines with empty numeric columns;
https://sources.debian.org/src/glibc/2.42-13/nss/fgetpwent_r.c#L57-L58

```c
	 INT_FIELD (result->pw_uid, ISCOLON, 0, 10,)
	 INT_FIELD (result->pw_gid, ISCOLON, 0, 10,)
```

Except for `+` special case?
https://sources.debian.org/src/glibc/2.42-13/nss/fgetpwent_r.c#L33-L46

```c
 STRING_FIELD (result->pw_name, ISCOLON, 0);
 if (line[0] == '\0'
     && (result->pw_name[0] == '+' || result->pw_name[0] == '-'))
   {
     /* This a special case.  We allow lines containing only a `+' sign
	since this is used for nss_compat.  All other services will
	reject this entry later.  Initialize all other fields now.  */
     result->pw_passwd = NULL;
     result->pw_uid = 0;
     result->pw_gid = 0;
     result->pw_gecos = NULL;
     result->pw_dir = NULL;
     result->pw_shell = NULL;
   }
```

thaJeztah:
Quick AI summary;

-----


- **Empty lines** are *not* parse errors. They are recognized before parsing and skipped.
- **Comment lines** are also recognized before parsing and skipped. The comment syntax is simply a line whose first non-NUL character is `#` (i.e., the first character on the line is `#`; there is no support for trailing/end-of-line comments in passwd/group files).
- **Malformed records** *do* cause the parser to report a parse failure (`EINVAL` internally), but that failure is treated as "this record is invalid", not "the file cannot be read". The iterator discards the record and continues with the next line.
- **Numeric fields** (UID/GID for normal passwd entries) must contain valid integers. Empty or non-numeric values are considered malformed.
- **String fields** may be empty. For example, an empty GECOS, home directory, or shell is perfectly valid.

So the overall behavior is:

| Input | Result |
|-------|--------|
| Empty line | skipped |
| `#` comment | skipped |
| Valid record | returned |
| Malformed record (missing required separators, bad UID/GID, etc.) | skipped |
| I/O error / scanner error | returned to caller |

That's a fairly common pattern in libc: the iteration APIs (`fgetpwent_r`, `fgetgrent_r`, etc.) are designed to enumerate **valid** records from what may be a partially-corrupt file. A malformed line is treated as a bad record, not as a fatal error for the entire database.

I think that's a strong precedent for `moby/sys`: if the library is exposing "iterate over passwd/group records", then **skipping malformed records without aborting the whole scan** is consistent with glibc's behavior. The only remaining design question is *what constitutes malformed*—and I'd argue empty numeric fields should be included in that category if the goal is to match glibc closely.

thaJeztah:
And for the `+` / `-` special case;

----

### `+` / `-` (NIS) entries

One exception to the above is the historical NIS ("Yellow Pages") extension, where passwd entries starting with `+` or `-` are **not normal passwd records**, but directives for the NSS/NIS compatibility layer.

Examples:

```text
+::::::::       # include all NIS users
+jane:::::::    # include one NIS user
-sam:::::::     # exclude one NIS user
```

glibc explicitly special-cases these records:

- **Normal passwd entries** (`root`, `adm`, ...):
  - UID/GID **must** contain valid integers.
  - Empty or invalid numeric fields → malformed → skipped.

- **NIS `+` / `-` entries**:
  - UID/GID **may** be empty.
  - Empty means "unspecified", not `0`.

This exists solely for compatibility with the legacy NIS passwd format.

For `moby/sys`, it may be reasonable **not to implement this compatibility extension**. The library is parsing flat passwd/group files rather than implementing glibc's full NSS stack, and NIS is effectively obsolete on modern Linux systems. In that case, `+`/`-` entries could simply be treated as unsupported/malformed records and skipped.

user/user.go
```diff
@@ -56,32 +56,49 @@ type IDMap struct {
...
 		// Use the type of the argument to figure out how to parse it, scanf() style.
 		// This is legit.
-		switch e := v[i].(type) {
+		switch e := v.(type) {
 		case *string:
 			*e = string(p)
 		case *int:
-			// "numbers", with conversion errors ignored because of some misbehaving configuration files.
-			*e, _ = strconv.Atoi(string(p))
+			n, ok, err := parseNumeric(string(p))
```
thaJeztah, line 88: note to self; this should not use `parseNumeric`, because it's too permissive. Perhaps we should rename it to `parseMaybeNumeric` to be more clear on intent (it's used for usernames that may be numeric); https://github.com/moby/sys/blob/4a8455d1645b3d4e6c0c652a1f45f384a1f1940f/user/user.go#L245-L253

PR #231, opened 2026-06-30
user: GetExecUser: rewrite with parseUserArg, parseGroupArg

thaJeztah:
Use a similar approach as was applied to GetAdditionalGroups in 6eb9f1533f265128f3172f26319ac4ec337e2cdc

### DIFF
--- a/user/user.go
+++ b/user/user.go
@@ -408,22 +408,6 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 	return user, nil
 }
 
-// groupArg is a parsed group argument for [GetAdditionalGroups].
-type groupArg struct {
-	name      string
-	gid       int
-	isNumeric bool
-}
-
-// matches reports whether group g satisfies the argument. Numeric arguments
-// are matched by GID only, others by name.
-func (ag groupArg) matches(g Group) bool {
-	if ag.isNumeric {
-		return g.Gid == ag.gid
-	}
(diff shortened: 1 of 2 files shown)


user/user.go
```diff
@@ -408,13 +393,59 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
...
+	name      string
+	uid       int
+	isNumeric bool
+}
+
+// parseUserArg parses a user argument as either a user name or UID.
+//
+// If name is empty, parseUserArg returns "nil, nil" to indicate that no user
+// argument was specified.
+func parseUserArg(name string) (*userArg, error) {
```
thaJeztah: Considering if we should move all these utilities to `user_utils.go`

vvoland (approved):

user/user.go
```diff
@@ -301,40 +301,35 @@ func GetExecUser(userSpec string, defaults *ExecUser, passwd, group io.Reader) (
 	}
 
 	// Allow for userSpec to have either "user", or optionally "user:group" syntax.
-	userArg, groupArg, _ := strings.Cut(userSpec, ":")
+	usr, grp, _ := strings.Cut(userSpec, ":")
 
 	// Convert userArg and groupArg to be numeric, so we don't have to execute
```
kolyshkin, line 306: This comment now refers to the old (before rename) variables.

PR #232, opened 2026-07-10
user: followups to #231

kolyshkin:
See commits for details.

It also might make sense to move `parseNumeric` to `user_utils.go` but I left it as is for now.


thaJeztah (approved):
LGTM, thx!

PR #235, opened 2026-07-21
parseNumeric: require non-empty values

thaJeztah:
follow-up to 841c83ee0d9315e5f4379361afe493c7e020c8c8 (https://github.com/moby/sys/pull/231) and 743614ec8aec7c62839dd3acf615da4e4de560d6 (https://github.com/moby/sys/pull/232).

Now that optional user-values are handled through parseGroupArg and parseUserArg (both of which gracefully handle empty values), we can make parseNumeric strict, and produce an error when trying to parse an empty string.


vvoland (approved):

kolyshkin (commented):
[post-merge] LGTM

PR #236, opened 2026-07-23
user: align Windows mkdir behavior with Unix

thaJeztah:
- relates to https://github.com/moby/moby/pull/16947

The Windows implementation of mkdirAs was originally introduced as a minimal wrapper around os.MkdirAll because ownership is not supported on Windows and the uid/gid parameters were only needed to keep the daemon starting.

As a result, MkdirAndChown would also create missing parent directories, unlike the Unix implementation.

Use os.Mkdir when mkAll is false to preserve the documented semantics, while retaining the simplified Windows implementation for ownership and permissions.

Originally introduced in [moby/moby@bfe252b78184][1], which added a Windows-specific implementation to work around limitations on that platform;

> A recent change to use pkg\idtools causes a chown to be on the
> startup path of the daemon. Chown is not supported on Windows,
> hence the daemon would not start.

[1]: https://github.com/moby/moby/commit/bfe252b78184d22d74a555a82aafc9e6dae3babc

### DIFF
--- a/user/idtools_windows.go
+++ b/user/idtools_windows.go
@@ -1,13 +1,17 @@
 package user
 
-import (
-	"os"
-)
+import "os"
 
-// This is currently a wrapper around [os.MkdirAll] since currently
-// permissions aren't set through this path, the identity isn't utilized.
-// Ownership is handled elsewhere, but in the future could be support here
-// too.
-func mkdirAs(path string, _ os.FileMode, _, _ int, _, _ bool) error {
-	return os.MkdirAll(path, 0)
+// mkdirAs creates path, optionally creating any missing parent directories.
+//
+// On Windows this is currently a thin wrapper around os.Mkdir and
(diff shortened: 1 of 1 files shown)


kolyshkin:
@thaJeztah looks like golangci-lint is not too happy with the changes

thaJeztah:
> The initial Stat and this Mkdir are not atomic: if another process creates path after the stat reports it missing, os.Mkdir returns EEXIST even though path is now a directory. That violates MkdirAndChown's documented nil-on-existing-directory behavior and is a regression from MkdirAll's race handling. Attempt Mkdir first and re-stat its error to accept a directory (while retaining the ENOTDIR result for files).


Had this initially, but then changed it to use the same order as on Linux; we can look at the linux code if we want to do a second check on errors, but that can be done separately.

thaJeztah:
cc @kolyshkin @vvoland ptal

Issue #239, opened 2026-07-27
userns.RunningInUserNS() misdetects systemd PrivateUsers=full as the initial user namespace


aleivag:
Hi maintainers,

`github.com/moby/sys/userns.RunningInUserNS()` currently detects the initial user namespace by parsing `/proc/self/uid_map`:

```go
// userns_linux.go
func uidMapInUserNS(uidMap string) bool {
    ...
    // As per user_namespaces(7), /proc/self/uid_map of
    // the initial user namespace shows 0 0 4294967295.
    initNS := a == 0 && b == 0 && c == 4294967295
    return !initNS
}
```

That assumes the only namespace that can contain 4294967295 UIDs is the initial namespace. However, uid_map describes ID mapping, not namespace identity. A non-initial user namespace can have a full identity map.

This now matters with systemd >= 260. (from: https://github.com/systemd/systemd/commit/24c73c79204bd22d243430bc0d9c2a7d814250bf) A service using `PrivateUsers=full` can run in a non-initial user namespace while still having a full identity UID/GID map:

```text
0 0 4294967295
```

In that case, `RunningInUserNS()` returns false even though the process is in a user namespace.

We hit this through runc, which vendors `moby/sys/userns`. runc's cgroup-v2 device controller intentionally skips/ignores some eBPF device-controller failures when running in a user namespace:

```go
// vendor/github.com/opencontainers/cgroups/devices/v2.go
func canSkipEBPFError(r *cgroups.Resources) bool {
    if userns.RunningInUserNS() {
        return true
    }
    ...
}
```

Because `RunningInUserNS()` returns false under `PrivateUsers=full`, runc does not take that userns path and instead fails nested container startup with:

```text
docker: Error response from daemon: failed to create task for container:
failed to create shim task: OCI runtime create failed: runc create failed:
unable to start container process: error during container init:
error setting cgroup config for procHooks process:
bpf_prog_query(BPF_CGROUP_DEVICE) failed: operation not permitted
```

...
The more robust check seems to be comparing `/proc/self/ns/user` to the kernel's initial user namespace inode, `PROC_USER_INIT_INO` (`0xEFFFFFFD`), rather than parsing `/proc/self/uid_map`. This is what systemd does, and what i propose we do in crun in https://github.com/containers/crun/issues/2150

Would you accept a patch changing `RunningInUserNS()` to use the namespace inode check instead of the `uid_map` full-range heuristic?

thaJeztah:
[...]

PR #240, opened 2026-07-28
userns: detect initial namespace by inode

aleivag:
The current uid_map heuristic treats a private user namespace with an identity uid_map as the initial user namespace. systemd 260 can create that shape with PrivateUsers=full, causing callers to believe they can perform initial-namespace-only operations such as cgroup device BPF setup.

Use /proc/self/ns/user's inode instead and compare it with the kernel's PROC_USER_INIT_INO value. This detects the initial user namespace directly instead of inferring it from the uid_map layout.

Related: https://github.com/moby/sys/issues/239

### DIFF
--- a/userns/userns_linux.go
+++ b/userns/userns_linux.go
@@ -5,7 +5,11 @@ import (
 	"fmt"
 	"os"
 	"sync"
+	"syscall"
 )
+
+// See PROC_USER_INIT_INO in https://github.com/torvalds/linux/blob/v7.1/include/uapi/linux/nsfs.h#L50.
+const procUserInitIno = 0xEFFFFFFD
 
 var inUserNS = sync.OnceValue(runningInUserNS)
 
@@ -17,6 +21,20 @@ var inUserNS = sync.OnceValue(runningInUserNS)
 // [libcontainer/runc]: https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/libcontainer/userns/userns_linux.go#L12-L49
 // [lcx/incus]: https://github.com/lxc/incus/blob/e45085dd42f826b3c8c3228e9733c0b6f998eafe/shared/util.go#L678-L700
 func runningInUserNS() bool {
+	var st syscall.Stat_t
+	if err := syscall.Stat("/proc/self/ns/user", &st); err == nil {
+		return st.Ino != procUserInitIno
+	} else if !os.IsNotExist(err) {
+		// As long as /proc/self/ns/user exists, we are on a modern kernel.
+		// Other errors indicate an unexpected procfs state, where assuming the
+		// init namespace would be unsafe.
+		return false
+	}
+
+	// Only fall back for older kernels that do not expose the user namespace
+	// through procfs at /proc/self/ns/user.
+	// TODO: Remove this fallback once Linux kernels older than 3.8 are no
+	// longer supported.
 	file, err := os.Open("/proc/self/uid_map")
 	if err != nil {
 		// This kernel-provided file only exists if user namespaces are supported.


userns/userns_linux.go
```diff
@@ -1,12 +1,12 @@
...
 
 import (
-	"bufio"
-	"fmt"
-	"os"
 	"sync"
+	"syscall"
 )
 
+const procUserInitIno = 0xEFFFFFFD
```
AkihiroSuda, line 12: Could you add a comment line to refer to https://github.com/torvalds/linux/blob/v7.1/include/uapi/linux/nsfs.h#L50

AkihiroSuda (approved):

userns/userns_linux.go
```diff
@@ -1,12 +1,13 @@
...
 
 import (
-	"bufio"
-	"fmt"
-	"os"
 	"sync"
+	"syscall"
 )
 
+// See PROC_USER_INIT_INO in include/uapi/linux/nsfs.h.
```
AkihiroSuda: full URL please

kolyshkin (commented):
1. I think that `TestInodeInUserNS` (and a separate `func inodeInUserNS`) makes little sense now when the code is very simple and clean. I mean, we just do stat(2) and compare `st.Ino` -- this needs no unit tests.

2. The problem is, this (predefined inode number for `PROC_USER_INIT_INO`) only appears in kernel v3.8 (it comes from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=98f842e675f9), and since Go itself (since Go 1.24) requires kernel v3.2 (see https://go.dev/wiki/MinimumRequirements#linuxlinux), we do need a fallback to old code. It will still work since most people run linux >= v3.8, and no one sane will run systemd v260+ on an older kernel.

kolyshkin:
> since Go itself (since Go 1.24) requires kernel v3.2

Note this package (github.com/moby/sys/userns) only requires Go 1.18 which itself has a lower linux kernel version requirement. Meaning, we definitely need the fallback.

kolyshkin (changes requested):
See my comments above

aleivag:
> Meaning, we definitely need the fallback.

cool... I went with "if /proc/self/ns/user exists use that one, if not, continue the old path" ... this should make the code more robust on new kernels, and , like you said, on on old kernels it matter less ..

kolyshkin:
> > Meaning, we definitely need the fallback.
> 
> cool... I went with "if /proc/self/ns/user exists use that one, if not, continue the old path" ... this should make the code more robust on new kernels, and , like you said, on on old kernels it matter less ..

OK, the `/proc/self/ns/user` appeared in v3.8 as well so it's OK to only fallback if the file itself does not exist.

Two notes though:
  - add a TODO to remove the fallback once kernel < 3.8 is no longer supported (fingers crossed it will happen in a few years);
  - keep the existing unit test intact, to minimize the change.

userns/userns_linux_test.go
AkihiroSuda, line 1: This test should be kept

userns/userns_linux.go
```diff
@@ -17,6 +21,15 @@ var inUserNS = sync.OnceValue(runningInUserNS)
 // [libcontainer/runc]: https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/libcontainer/userns/userns_linux.go#L12-L49
 // [lcx/incus]: https://github.com/lxc/incus/blob/e45085dd42f826b3c8c3228e9733c0b6f998eafe/shared/util.go#L678-L700
 func runningInUserNS() bool {
+	var st syscall.Stat_t
+	if err := syscall.Stat("/proc/self/ns/user", &st); err == nil {
+		return st.Ino != procUserInitIno
+	} else if !os.IsNotExist(err) {
+		return false
```
AkihiroSuda, line 31: This should still fall back to the legacy method?

aleivag: > This should still fall back to the legacy method?

I can do this... but i think this might be the wrong way of doing it...  consider thisL If the file "/proc/self/ns/user" is there, it means a modern kernel, so if stat fails for any other reason than "file does not exists" it mens that you are still on a modern kernel but something unexpected is wrong with /proc, permissions, mount state, or namespace access, its safer to assume that you are not in the init namespace than to fallback in the old criteria (specially if you plan to remove this line once older kernel are not supported) ...

aleivag:
@AkihiroSuda @kolyshkin may i bug you with one more review.  ppllleeeeeeasseee? 😃

AkihiroSuda (approved):

userns/userns_linux.go
thaJeztah: I agree with this logic; the only thing i'm wondering is; do we need a comment here to describe the logic?

Not a blocker, just putting it on the table; a future visitor of the code may think we either accidentally fall through, or return too early.

thaJeztah (approved):
LGTM

left a comment https://github.com/moby/sys/pull/240#discussion_r3719328483, but not a blocker; @kolyshkin can you PTAL?

aleivag:
> LGTM
> 
> left a comment [#240 (comment)](https://github.com/moby/sys/pull/240#discussion_r3719328483), but not a blocker; 

@thaJeztah updated with the comment... @kolyshkin its up to you now! 😄

aleivag:
mr @kolyshkin can i bug you with one more release please ?

kolyshkin (approved):
LGTM, thanks!

(not sure what happened to our CI here)

kolyshkin:
Let me close/reopen this one to kick CI

aleivag:
Thank you All !!

PR #245, opened 2026-08-07
enhance linters configuration and improve code readability

mmorel-35:
This introduces several improvements and code quality enhancements across the codebase, including updates to linter configuration, refactoring for clarity and maintainability, and minor bug fixes. The most significant changes are grouped below.

**Linter and Code Quality Enhancements:**

* `.golangci.yml`: Enabled additional linters (`gocritic`, `modernize`, `revive`) and configured their rules to catch more code issues and enforce consistent style. Also, set `max-issues-per-linter` and `max-same-issues` to 0 to report all issues. 
* Refactored several test and function signatures to use blank identifiers for unused parameters, improving code clarity and linter compliance .

**Code Refactoring and Simplification:**

* `user/idtools_unix.go`: Refactored variable assignments to use short variable declarations for error handling, improving readability and preventing potential variable shadowing issues.
* `user/idtools_unix.go`: Simplified increment operation in `lookupSubRangesFile` for clarity.

**Bug Fixes and Test Improvements:**

* `capability/capability_linux.go`: Replaced `if/else` chains with a `switch` statement in `mkString` for improved logic clarity.
* `capability/capability_test.go`: Improved empty string check for bounding set in tests.
* `mount/sharedsubtree_linux_test.go`: Removed unnecessary deferred function in `TestSubtreeUnbindable` for cleanup, streamlining test logic.

**Minor Corrections and Consistency Improvements:**

* `signal/signal.go`: Renamed a variable from `signal` to `sig` to avoid shadowing and improve code clarity.
* `mount/mounter_linux_test.go`: Updated function signature for `validateMount` to use a more concise parameter list.
* `mountinfo/mounted_linux_test.go` Improved test logic for bind mount detection and error handling.

PR #246, opened 2026-08-11
chore: enable and fixes rules from gocritic

mmorel-35:
Enables all gocritic rules by default, then excludes the one non compliant then fix one rule per commit
Covers #245 gocritic part only

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,7 +25,6 @@ linters:
         - rangeValCopy
         - sloppyReassign
         - unnamedResult
-        - unnecessaryDefer
         - whyNoLint
       enable-all: true
     staticcheck:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -19,7 +19,6 @@ linters:
         - builtinShadow
         - deferInLoop
         - hugeParam
-        - ifElseChain
         - importShadow
         - paramTypeCombine
         - rangeValCopy
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,7 +18,6 @@ linters:
         - assignOp
         - builtinShadow
         - deferInLoop
-        - emptyStringTest
         - hugeParam
         - ifElseChain
         - importShadow
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,7 +15,6 @@ linters:
     gocritic:
       disabled-checks:
         - appendAssign
-        - assignOp
         - builtinShadow
         - deferInLoop
         - hugeParam
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,6 +3,7 @@ version: "2"
 linters:
   enable:
     - errorlint
+    - gocritic
     - unconvert
     - unparam
   exclusions:
@@ -11,6 +12,23 @@ linters:
       - comments
       - std-error-handling
   settings:
+    gocritic:
+      disabled-checks:
+        - appendAssign
+        - assignOp
+        - builtinShadow
+        - deferInLoop
+        - emptyStringTest
+        - hugeParam
+        - ifElseChain
+        - importShadow
+        - paramTypeCombine
+        - rangeValCopy
+        - sloppyReassign
+        - unnamedResult
+        - unnecessaryDefer
+        - whyNoLint
+      enable-all: true
     staticcheck:
       # Enable all options, with some exceptions.
       # For defaults, see https://golangci-lint.run/usage/linters/#staticcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,7 +18,6 @@ linters:
         - deferInLoop
         - hugeParam
         - importShadow
-        - paramTypeCombine
         - rangeValCopy
         - sloppyReassign
         - unnamedResult
@@ -36,3 +35,7 @@ formatters:
     - gofumpt
   exclusions:
     generated: disable
+  settings:
+    gofumpt:
+      extra:
+        group-params: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,7 +15,6 @@ linters:
     gocritic:
       disabled-checks:
         - appendAssign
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -36,8 +36,4 @@ formatters:
   enable:
     - gofumpt
   exclusions:
-    generated: disable
-
-issues:
-  max-issues-per-linter: 0
-  max-same-issues: 0
+    generated: disable
```
kolyshkin: nit: please don't remove `\n` from the last line.

kolyshkin: IOW please squash the two last commits

.golangci.yml
```diff
@@ -36,8 +36,4 @@ formatters:
   enable:
     - gofumpt
   exclusions:
-    generated: disable
-
-issues:
-  max-issues-per-linter: 0
-  max-same-issues: 0
```
kolyshkin: So, in the very same PR you add those rules then remove them. Feels like unnecessary noise to me -- why skip adding it in the first place?

kolyshkin:
Overall LGTM, left some nits.

.golangci.yml
mmorel-35: It's merged in the first commit

.golangci.yml
mmorel-35: I first thought enabling it then changed my mind. It allows to see all issues instead of 3 per rule at a time. So teams don't like it. What do you think ?

ccoVeille (approved):

.golangci.yml
```diff
@@ -11,6 +12,19 @@ linters:
...
+        - appendAssign
+        - builtinShadow
+        - deferInLoop
+        - hugeParam
+        - importShadow
+        - paramTypeCombine
+        - rangeValCopy
+        - sloppyReassign
+        - unnamedResult
+        - whyNoLint
```
thaJeztah: Is the intent to address all of these (in follow-ups), or are there that cannot (yet) be addressed (e.g. due to go versions).

It's good to leave a comment for each to outline if it's either a permanent (won't fix because) or temporary (TODO).

At least some of them stood out as "should likely be fixed" (`whyNoLint`, `rangeValCopy`, `importShadow`), and I know the `builtinShadow` is a bit too opinionated (it's "OK", but sometimes not worth the code-churn.

mmorel-35: Do you already know if you don't want some of them ? Otherwise the intention is indeed to try to comply with those that are the moby group and the sys project particularly usual practices.

kolyshkin (approved):
LGTM; @thaJeztah PTAL

thaJeztah:
Oh! You need to rebase, because go1.27 is not "stable", so GolangCi-lint needed updating

.golangci.yml
thaJeztah: I was curious what it would flag, and overall it's not too problematic in this repository to enable them all; the one I was most on the fence about was `unnamedResults`; 

- https://github.com/moby/sys/pull/252


BUT I also ran this branch, and looks like some were not hit, so probably could already be removed from the exclusions list (we could always add an exclusion if we encounter one that we don't think should be fixed); so if you can update to this list;

```yaml
      disabled-checks:
        - appendAssign
        - hugeParam
        - importShadow
        - rangeValCopy
        - sloppyReassign
        - unnamedResult
```

(or fix the trivial ones, but we can do that in my follow-up)

thaJeztah (changes requested):
needs a rebase (see above), but LGTM after the list is updated to the ones we actually hit; https://github.com/moby/sys/pull/246#discussion_r3932231012

mount/mount_unix_test.go
```diff
@@ -232,6 +224,18 @@ func TestRecursiveUnmountTooGreedy(t *testing.T) {
...
+func createAndMountTmpfs(t *testing.T, tmp, d string) {
+	dir := path.Join(tmp, d)
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		t.Fatal(err)
+	}
+	if err := Mount("tmpfs", dir, "tmpfs", ""); err != nil {
+		t.Fatal(err)
+	}
+	//nolint:errcheck // Ignore error in tests
+	Unmount(dir)
```
thaJeztah, line 236: Perhaps just explicitly ignore the error, then we don't need a nolint comment;



```suggestion
	_ = Unmount(dir)
```

thaJeztah: I also don't think moving this to a utility is correct; it changes semantics because it now unmounts immediately?  If we want to keep that behavior, then we could do something like;

```go
var toUnMount []string

t.Cleanup(func() {
    for _ m := range toUnmount {
        _ = Unmount(m)
   }
})
```

(either `t.Cleanup` or `defer`, depending on go version supported, but I'd be fine bumping to a slightly newer version if needed)

mount/mount_unix_test.go
```diff
@@ -273,3 +273,10 @@ func TestRecursiveUnmount_SubMountFailsToUnmount(t *testing.T) {
 		})
 	}
 }
+
+func mountTemporaryDirectory(t *testing.T, dir string) {
+	if err := Mount("tmpfs", dir, "tmpfs", ""); err != nil {
+		t.Fatal(err)
+	}
+	Unmount(dir) //nolint:errcheck // Ignore error in tests
+}
```
thaJeztah, line 282: Same here; we can also defer the unmount explicitly ignore errors here as well;


```go
_ = Unmount(dir) //nolint:errcheck // Ignore error in tests
```

If the `defer` / cleanup wasn't relevant to the test, we can just inline it without the cleanup.

mount/mount_unix_test.go
thaJeztah: Possibly the utility of https://github.com/moby/sys/pull/253 could be used if it's modified to take a variadic strings.

PR #249, opened 2026-08-27
signal: add SignalMap for linux/sparc64

shalseth:
SPARC has no SIGSTKFLT, so x/sys/unix does not define it for sparc64 and the "STKFLT" entry in signal_linux.go does not compile there. SPARC uses SIGEMT in that slot, as the mips architectures do.

Add signal_linux_sparc64.go and exclude sparc64 from signal_linux.go, following signal_linux_mipsx.go. The new file is that one with the build tag changed and sigrtmax set to 64 rather than 127, which leaves sigrtmin and sigrtmax matching the generic file, so the SIGSTKFLT/SIGEMT swap is the only substantive difference.

One thing is still missing before the package builds on linux/sparc64: x/sys/unix has no syscall implementation for that architecture. That is a separate matter, being pursued upstream. This change was checked against a tree that has one, where the package's tests pass on an UltraSPARC T4-1.


kolyshkin (changes requested):
I think we should put this on hold until x/sys/unix implements linux/sparc64 syscalls.

shalseth:
> I think we should put this on hold until x/sys/unix implements linux/sparc64 syscalls.

Agreed, posted now on https://github.com/golang/sys/pull/292.
Had some other x/sys PR's pending approval so held it, but looks like they are all getting merged now.

kolyshkin:
Draft pending https://go-review.googlesource.com/c/sys/+/823604 merge

shalseth:
> Draft pending https://go-review.googlesource.com/c/sys/+/823604 merge

Now merged

PR #256, opened 2026-09-08
user: fall back to getent when /etc/passwd lookup fails

dims:
## What this does

`LoadIdentityMapping` resolves the requested user through `LookupUser`, which only reads `/etc/passwd`. Accounts that come from NSS (LDAP, sssd, systemd-userdb) are invisible to it, even when they work everywhere else on the host. This closes the `TODO: Consider adding support for calling out to "getent"` that's been in the function since it was extracted from Docker/runc.

## The fix

When `LookupUser` returns `ErrNoPasswdEntries`, fall back to `getent passwd <name>`, which does consult NSS. The output is parsed with the package's existing `ParsePasswdFilter`, so there's no new parsing logic. Any other error from `LookupUser` is returned unchanged. If `getent` isn't installed or also can't find the account, behavior is identical to today.

## Motivation

Kubernetes hit this exact gap: kubelet's move to `CGO_ENABLED=0` (kubernetes/kubernetes#135870) broke user-namespace ID range lookups for NSS-backed `kubelet` accounts, because Go's pure-Go `os/user` has the same `/etc/passwd`-only limitation (tracked upstream as golang/go#38810, open since 2020). This library already isolates the "resolve a user" step behind `LookupUser`, making it a lower-risk place to close the gap.

See:
- https://go-review.googlesource.com/c/go/+/785361
- https://go-review.googlesource.com/c/go/+/459455

## Testing

- `go build`, `go vet`, `gofmt -l` clean.
- New `TestLookupUserViaGetent` covers both branches (found via getent, not found), gated on `getent` being on `PATH`. Verified passing against the real binary on Linux; skips cleanly where unavailable (macOS, Windows).
- Full existing `user` package test suite still passes.


kolyshkin (commented):
Interestingly, `/etc/subuid` and `/etc/subgid` can have either name or uid as the first field, and `LoadIdentityMapping` accepts `name`. Given that we don't use `uid` anywhere else I _think_ we can fall back to using `name` directly to get the subuid/subgid.

IOW something like:
```go
usr, err := LookupUser(name)
if err != nil {
  usr, err = lookupUserViaGetent(name)
  if err != nil {
     // Fall back to name lookup.
     usr = User{Name: name, Uid: -1}
  }
}
```

Maybe we can skip `getent` entirely and just look up by `name`?

PR #258, opened 2026-09-09
Makefile: fix testing and building of subpackages

thaJeztah:
- relates to https://github.com/moby/sys/pull/211#discussion_r3962360520

The Makefile iterates over modules, but its test and cross-build commands only operated on the package at each module root.

Use `./...` so subpackages are included in tests, local dependency tests, and cross-platform builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@ clean:
 	$(RM) */coverage.txt
 
 .PHONY: foreach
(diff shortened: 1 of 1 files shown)


vvoland (approved):

kolyshkin (commented):
post-merge LGTM

